PR #1, opened 2023-06-14
added user modal and updated user-profile user fetching method

UncleBu1728:
added a model for new users to input names to avoid undefined name column
updated user-profile method to get user data

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -1,4 +1,6 @@
 import {
+  Alert,
+  AlertIcon,
   Avatar,
   Box,
   Button,
@@ -6,13 +8,21 @@ import {
   CardBody,
   CardHeader,
   Flex,
+  Modal,
+  ModalOverlay,
+  ModalContent,
+  ModalHeader,
+  ModalBody,
   Heading,
   StackDivider,
   Text,
   VStack,
   useColorModeValue,
+  FormLabel,
+  Input,
+  FormControl,
 } from '@chakra-ui/react';
-import React from 'react'
+import React, { useState } from 'react';
 import { NextPageWithLayout } from "../NextPageWithLayout";
 import AppLayout from "../layouts";
 import useUserInfo from "../useUserInfo";
@@ -21,15 +31,102 @@ import tClientNext from "../tClientNext";
 import PublicUser from '../shared/publicModels/PublicUser';
 import PublicGroup from '../shared/publicModels/PublicGroup';
 import { MdVideocam } from 'react-icons/md';
+import { HSeparator } from 'horizon-ui/components/separator/Separator';
+import { IYuanjianUser } from 'shared/user';
+import { toast } from "react-toastify";
 
 const AppIndex: NextPageWithLayout = () => {
   const user = useUserInfo();
+  if(!user.name){
+    return <Box paddingTop={'80px'}><SetName u = {user} /><Meetings /></Box>
+  };
   return <Box paddingTop={'80px'}><Meetings /></Box>;
 }
 
 AppIndex.getLayout = (page) => <AppLayout>{page}</AppLayout>;
 
 export default AppIndex;
+
+function SetName({u} : {u:IYuanjianUser}) {
+  const textColor = useColorModeValue('navy.700', 'white');
+  const brandStars = useColorModeValue('brand.500', 'brand.400');
+  const [isOpen, setOpen] = useState(true);
+  const [name, setName] = useState('');
+  
+  const handleSubmit = async () => {
+    if (name) {
+      const updatedUser: IYuanjianUser = {
+        id: u.id,
+        pinyin: name,
+        name: name,
+        email: u.email,
+        roles: u.roles,
+        clientId: u.clientId,
+      };
+
+      tClientBrowser.user.updateProfile.mutate(updatedUser).then(
+        res => {
+          if (res === "ok") {
+            console.log("user update succeeded");
+            setOpen(false);     
+          }
+        }
+      ).catch(e => toast.error(e.message, { autoClose: false }))
+    };
+  };
+
+  return (
+    <Modal isOpen={isOpen} onClose={() => setOpen(false)}>
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>新用户登录</ModalHeader>
+        <ModalBody>
+          <Text>
+            请填写个人中文全名
+          </Text>
+          <Box mt={4}>
+            <Flex align='center' mb='25px'>
+              <HSeparator />
+            </Flex>
+            <FormControl>
+              <FormLabel display='flex' ms='4px' fontSize='md' fontWeight='500' color={textColor} mb='8px'>
+                Email: <Text ml={4} fontSize='md' fontWeight='900' color={textColor}>{u.email}</Text>
+              </FormLabel>
+              {!name && (
+                <Alert status="error" mt={4}>
+                  <AlertIcon />
+                  用户姓名不能为空
+                </Alert>
+              )}
+              <FormLabel display='flex' ms='4px' fontSize='md' fontWeight='500' color={textColor} mb='8px'>
+                姓名<Text color={brandStars}>*</Text>
+              </FormLabel>
+              <Input
+                isRequired={true}
+                value={name}
+                onChange={(e) => setName(e.target.value)}
+                variant='auth'
+                fontSize='sm'
+                ms={{ base: '0px', md: '0px' }}
+                type='email'
+                placeholder='张三'
+                mb='24px'
+                fontWeight='500'
+                size='lg'
+              />
+              <Button
+                onClick={handleSubmit}
+                loadingText='Creating'
+                fontSize='sm' variant='brand' fontWeight='500' w='100%' h='50' mb='24px'>
+                确认提交
+              </Button>
+            </FormControl>
+          </Box>
+        </ModalBody>
+      </ModalContent>
+    </Modal>
+  );
+}
 
 function Meetings() {
   const { data } = tClientNext.myMeetings.list.useQuery({});
@@ -42,7 +139,7 @@ function Meetings() {
       </CardHeader>
       <CardBody>
         <VStack divider={<StackDivider />} align='left' spacing='6'>
-          {data && 
+          {data &&
             data.groupList.map((group: PublicGroup, idx: any) => Meeting(myUserId, group, data.userMap))
           }
         </VStack>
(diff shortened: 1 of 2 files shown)


src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
 import PublicUser from '../shared/publicModels/PublicUser';
 import PublicGroup from '../shared/publicModels/PublicGroup';
 import { MdVideocam } from 'react-icons/md';
+import { HSeparator } from 'horizon-ui/components/separator/Separator';
+import { IYuanjianUser } from 'shared/user';
+import { toast } from "react-toastify";
 
 const AppIndex: NextPageWithLayout = () => {
   const user = useUserInfo();
+  if(!user.name){
```
weihanwang, line 40: spaces needed before ( and after ). alternatively, install formatter to automatically format changed portion of a source file on file saving.

src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
...
+    return <Box paddingTop={'80px'}><SetName u = {user} /><Meetings /></Box>
+  };
   return <Box paddingTop={'80px'}><Meetings /></Box>;
 }
 
 AppIndex.getLayout = (page) => <AppLayout>{page}</AppLayout>;
 
 export default AppIndex;
 
+function SetName({u} : {u:IYuanjianUser}) {
```
weihanwang, line 50: 1. `SetNameModal` to indicate the returned component type
2. add a space after :. please also check format for the rest of the PR.

src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
...
 import PublicGroup from '../shared/publicModels/PublicGroup';
 import { MdVideocam } from 'react-icons/md';
+import { HSeparator } from 'horizon-ui/components/separator/Separator';
+import { IYuanjianUser } from 'shared/user';
+import { toast } from "react-toastify";
 
 const AppIndex: NextPageWithLayout = () => {
   const user = useUserInfo();
+  if(!user.name){
+    return <Box paddingTop={'80px'}><SetName u = {user} /><Meetings /></Box>
```
weihanwang, line 41: do you need this box for modal?

src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
...
+  };
   return <Box paddingTop={'80px'}><Meetings /></Box>;
```
weihanwang, line 43: use a simpler format: `return user.name ? <SetNameModal ... /> : <Box ... />`

src/pages/index.tsx
weihanwang, line 40: did you verify that once the user name is set and the modal is closed, the ui is refreshed to show <Meetings> without the need to reload the entire page?

src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
...
+  const textColor = useColorModeValue('navy.700', 'white');
+  const brandStars = useColorModeValue('brand.500', 'brand.400');
+  const [isOpen, setOpen] = useState(true);
+  const [name, setName] = useState('');
+  
+  const handleSubmit = async () => {
+    if (name) {
+      const updatedUser: IYuanjianUser = {
+        id: u.id,
+        pinyin: name,
```
weihanwang, line 60: why name itself and not the pinyin of the name?

src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
...
+        name: name,
+        email: u.email,
+        roles: u.roles,
+        clientId: u.clientId,
```
weihanwang, line 64: why do we need to pass in these fields even if they are not updated?

src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
...
+      };
+
+      tClientBrowser.user.updateProfile.mutate(updatedUser).then(
+        res => {
+          if (res === "ok") {
+            console.log("user update succeeded");
```
weihanwang, line 70: "user **name** update succeeded"

src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
...
+            setOpen(false);     
+          }
+        }
+      ).catch(e => toast.error(e.message, { autoClose: false }))
+    };
+  };
+
+  return (
+    <Modal isOpen={isOpen} onClose={() => setOpen(false)}>
```
weihanwang, line 79: Why do we allow the user to close the dialog without entering a valid name?

src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
...
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>新用户登录</ModalHeader>
+        <ModalBody>
+          <Text>
+            请填写个人中文全名
```
weihanwang, line 85: remove "个人"

src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
...
+          </Text>
+          <Box mt={4}>
+            <Flex align='center' mb='25px'>
+              <HSeparator />
+            </Flex>
+            <FormControl>
+              <FormLabel display='flex' ms='4px' fontSize='md' fontWeight='500' color={textColor} mb='8px'>
```
weihanwang, line 92: let's remove styling tags as many as possible, to rely on default styling as much as possible.

src/pages/index.tsx
```diff
@@ -21,16 +31,103 @@ import tClientNext from "../tClientNext";
...
+                Email: <Text ml={4} fontSize='md' fontWeight='900' color={textColor}>{u.email}</Text>
+              </FormLabel>
+              {!name && (
+                <Alert status="error" mt={4}>
+                  <AlertIcon />
+                  用户姓名不能为空
```
weihanwang, line 98: remove "用户”

src/pages/index.tsx
weihanwang, line 92: We don't need to display email here.

weihanwang: the same for all other components below and elsewhere.

src/pages/index.tsx
```diff
@@ -58,19 +155,19 @@ function Meeting(myUserId: string, group: PublicGroup, userMap: Record<string, P
...
+          </>;
+        }
         )
       }
     </Flex>
   );
 }
 
 async function launchMeeting(groupId: string) {
-  const meetingLink = await tClientBrowser.myMeetings.generateMeetingLink.mutate({groupId: groupId});
```
weihanwang, line 74: is this the result of an auto formatter?

weihanwang (changes requested):

src/pages/index.tsx
UncleBu1728: ESlint installed

weihanwang: I don't think ESlink is primarily used for code formatting -- at least it didn't fix formatting issues with this PR :)

src/pages/index.tsx
UncleBu1728: The page will not re-render since the modal closed will not update any properties on the meeting page, so my solution was to force users that have not entered names to enter a page with the modal

src/pages/index.tsx
UncleBu1728: the return object was IYuanjianUser type. The type is removed and an object with only updated name is used now

src/pages/index.tsx
UncleBu1728: onClose is required for Modal, a function that return undefined is used now

src/pages/index.tsx
```diff
@@ -21,16 +31,94 @@ import tClientNext from "../tClientNext";
...
 import PublicGroup from '../shared/publicModels/PublicGroup';
 import { MdVideocam } from 'react-icons/md';
+import { HSeparator } from 'horizon-ui/components/separator/Separator';
+import { IYuanjianUser } from 'shared/user';
+import { toast } from "react-toastify";
 
 const AppIndex: NextPageWithLayout = () => {
   const user = useUserInfo();
-  return <Box paddingTop={'80px'}><Meetings /></Box>;
+  // if (!user.name) {
```
weihanwang, line 40: remove unused code (and not just comment out)

src/pages/index.tsx
```diff
@@ -21,16 +31,94 @@ import tClientNext from "../tClientNext";
...
+  //   return <Box paddingTop={'80px'}><SetNameModal u={user} /><Meetings /></Box>
+  // };
+  return <Box paddingTop={'80px'}> {user.name ? <></>:<SetNameModal />}<Meetings /></Box>
```
weihanwang, line 43: spaces before and after ":" and braces

src/pages/index.tsx
```diff
@@ -21,16 +31,94 @@ import tClientNext from "../tClientNext";
...
+            </Flex>
+            <FormControl>
+              {!name && (
+                <Alert status="error" mt={4}>
+                  <AlertIcon />
+                  姓名不能为空
+                </Alert>
+              )}
+              <FormLabel display='flex'>
+                姓名<Text color={brandStars}>*</Text>
```
weihanwang, line 93: I don't think we need the star here because it's easy for user to understand (from the UI message) that it's a required field.

src/pages/index.tsx
```diff
@@ -21,16 +31,94 @@ import tClientNext from "../tClientNext";
...
+              </FormLabel>
+              <Input
```
weihanwang, line 95: remove as many styling props from both <Input> and <Button>. do we really need them? Right now we don't need to (and shouldn't) fine tune local styles. We should adjust styling globally.

weihanwang (changes requested):

src/pages/user-profile.tsx
```diff
@@ -17,47 +17,37 @@ import tClientBrowser from "../tClientBrowser";
...
-  };
+    setName(user.name)
+  }, [user]);
 
   const handleSubmit = async () => {
     if (name) {
-      const updatedUser: IYuanjianUser = {
-        id: userProfile.id,
-        pinyin: name,
+      const nameUpdatedUser: IYuanjianUser = {
```
weihanwang, line 33: same here: 1. `updatedUser` is good. 2. clone the entire object instead.

src/pages/user-profile.tsx
```diff
@@ -17,47 +17,37 @@ import tClientBrowser from "../tClientBrowser";
...
           if (res === "ok") {
-            console.log("user update succeeded")
-            setShow(!show)
+            console.log("user name update succeeded");
+            updateUser(nameUpdatedUser);
+            setShow(!show);
           }
         }
-      ).catch(e => toast.error(e.message, { autoClose: false }))
+      ).catch(e => toast.error(e.message, { autoClose: false }));
```
weihanwang, line 50: why not autoclose?

weihanwang (changes requested):

weihanwang (approved):

PR #23, opened 2023-06-20
Added sendgrid feature

UncleBu1728:
Added SendGrid API to send emails and notify users
API key and Template keys required to change the placeholders
The API key is required to be added to the environment variables

### DIFF
--- a/src/api/routes/myMeetings.ts
+++ b/src/api/routes/myMeetings.ts
@@ -18,6 +18,7 @@ function isSubset<T>(superset: Set<T>, subset: Set<T>): boolean {
   }
   return true;
 }
+
 /*
 {
   "meeting_number": 1,
--- a/src/api/routes/user.ts
+++ b/src/api/routes/user.ts
@@ -17,9 +17,9 @@ const user = router({
     auth('profile:write')
   ).input(
     // A Chinese name must have at least 2 characters.
-    z.object({ name: z.string().min(2, "required")})
+    z.object({ name: z.string().min(2, "required") })
   ).mutation(async ({ input, ctx }) => {
-    await ctx.user.update({
+  await ctx.user.update({
       name: input.name,
       pinyin: pinyin.convertToPinyin(input.name),
     });
--- a/src/api/tencentMeeting.ts
+++ b/src/api/tencentMeeting.ts
@@ -1,25 +1,23 @@
 import crypto from 'crypto';
 import qs from 'qs';
 import apiEnv from "./apiEnv";
-import * as https from "https";
-import * as http from "http";
+import https from "https";
+import http from "http";
 import z from "zod";
+
+const LOG_HEADER = "[TecentMeeting]"
 
 const splitFirst = (s: string, separator: string) => {
   const idx = s.indexOf(separator);
-  if (idx < 0) {
-    return [s];
-  }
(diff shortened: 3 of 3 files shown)


src/api/auth.ts
```diff
@@ -88,6 +89,16 @@ async function findOrCreateUser(clientId: string, email: string): Promise<User>
     const roles: [Role] = [isAdmin ? 'ADMIN' : 'VISITOR'];
     console.log(`Creating user ${email} roles ${roles} id ${clientId}`);
     try {
+      // using sendgrid to notify admin a new user is created
```
weihanwang, line 92: 1. move this block to outside of and _before_ this try block. The try block is for User.create() only. Naturally email sending should be after user is created in db, but logically only the reverse order is error proof. (I can explain further if needed.)

2. create a new function in sendgrid.ts called sendEmailToAdmin() which only accept one parameter which is the message to be included in the email. So here you can simply call `sendEmailToAdmin(`New user ${email} signed up`)`

src/api/sendGrid.ts
```diff
@@ -0,0 +1,23 @@
+const sgMail = require('@sendgrid/mail');
```
weihanwang, line 1: 1. use `import` syntax
2. avoid acronyms when possible.`mail` is fine.

src/api/sendGrid.ts
```diff
@@ -0,0 +1,23 @@
...
+const template_id = 'd-2e3aabdb9a3c45a393877596534c57c2'; // temporary holder, change to required template id
```
weihanwang, line 2: move it into sendEmailToAdmin()

src/api/sendGrid.ts
weihanwang, line 1: sendgrid.ts (all lower case, to follow sendgrid's official package convention ("@sendgrid")).

src/api/sendGrid.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface Personalizations {
+    to: string|string[];
```
weihanwang, line 5: it should be { "to": [ { "email": "...", "name": "..." } ]

src/api/sendGrid.ts
```diff
@@ -0,0 +1,23 @@
...
+    dynamic_template_data?: Record<string, string>;
+}
+
+export const sendEmail = async (personalizations: Personalizations[]) => {
+
+    sgMail.setApiKey(process.env.SENDGRID_API_KEY);
```
weihanwang, line 11: 1. move it to outside of this function
2. don't call process.env directly. use apiEnv instead. refer to other callers of apiEnv.

src/api/sendGrid.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    const message = {
+        personalizations: personalizations,
+        from: 'yizhang1728@gmail.com', // yuanjian@163
+        template_id: template_id,
+    }
+
+    sgMail
```
weihanwang, line 19: avoid using `.then`. use await:
```
console.log(`Sending email: ${personalization}`)
await mail.send()
```

Let upper stacks catch the error. Do not "swallow" the error here at the lower stack -- this is an anti-pattern against "fail fast and fail safe".

weihanwang (changes requested):

PR #24, opened 2023-06-20
added sendgrid feature

UncleBu1728:
added SendGrid mail feature to allow sending emails via API key and template id
added SendEmailToAdmin to notify admins when new users registered 
**API key and Template id** needed to be replaced with proper ones before officially deployed


src/api/auth.ts
```diff
@@ -87,6 +88,8 @@ async function findOrCreateUser(clientId: string, email: string): Promise<User>
     const isAdmin = apiEnv.ASSIGN_ADMIN_ROLE_ON_SIGN_UP.includes(email);
     const roles: [Role] = [isAdmin ? 'ADMIN' : 'VISITOR'];
     console.log(`Creating user ${email} roles ${roles} id ${clientId}`);
+    // using sendgrid to notify admin a new user is created
```
weihanwang, line 91: This comment doesn't add value because the code is already self-explanatory (which is ideal).

src/api/sendgrid.ts
```diff
@@ -0,0 +1,53 @@
...
+
+mail.setApiKey(apiEnv.SENDGRID_API_KEY);
+
+const sendEmail = async (personalizations: Personalizations[], template_id: string) => {
+
+    console.log("SendGrid Mail Executed");
+
+    const message: MailDataRequired = {
+        personalizations: personalizations,
+        from: 'yizhang1728@gmail.com', // yuanjian@163
```
weihanwang, line 20: use a fake email. you don't want your real email address to live in git history forever.

src/api/sendgrid.ts
```diff
@@ -0,0 +1,53 @@
+import mail, { MailDataRequired } from '@sendgrid/mail'
+import apiEnv from './apiEnv';
+
+export interface Personalizations {
```
weihanwang, line 4: Why not use sendgrid's native definition of PersonalizationData? https://github.com/sendgrid/sendgrid-nodejs/blob/b8125d86134157a462c2e24816cfa56677ab76a8/packages/helpers/classes/personalization.d.ts#L3

src/api/sendgrid.ts
```diff
@@ -0,0 +1,53 @@
...
+    const message: MailDataRequired = {
+        personalizations: personalizations,
+        from: 'yizhang1728@gmail.com', // yuanjian@163
+        templateId: template_id,
+    };
+
+    try {
+        const res = await mail.send(message);
+        console.log("SendGrid Mail sent", res);
+    } catch (e) {
```
weihanwang, line 27: 1. don't try, don't catch, don't swallow exceptions. let it propagate to upper levels. simply:

```
await mail.send()
```

with nothing else.

2. do you need to handle the return value for error detection? if not, remove `const res =`.

src/api/sendgrid.ts
```diff
@@ -0,0 +1,53 @@
...
+        name: string
+    }];
+    dynamic_template_data?: Record<string, string>;
+}
+
+mail.setApiKey(apiEnv.SENDGRID_API_KEY);
+
+const sendEmail = async (personalizations: Personalizations[], template_id: string) => {
+
+    console.log("SendGrid Mail Executed");
```
weihanwang, line 16: it's not executed yet. print `Sending mail via SendGrid: ${JSON.stringify(personalizations, null, 2)}` 

also, move this line right above calling mail.send()

src/api/sendgrid.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    const message: MailDataRequired = {
+        personalizations: personalizations,
+        from: 'yizhang1728@gmail.com', // yuanjian@163
+        templateId: template_id,
+    };
+
+    try {
+        const res = await mail.send(message);
+        console.log("SendGrid Mail sent", res);
```
weihanwang, line 26: no need for this log line.

src/api/sendgrid.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+export const sendEmailToAdmin = async (new_user_email: string) => {
+    const template_id = 'd-2e3aabdb9a3c45a393877596534c57c2'; // temporary holder, change to required template id
+
+    const personalizations: Personalizations[] = [{
+        // change/add codes below to designated email addresses and names
+        to:
+            [
```
weihanwang, line 38: nit: compress the formatt a bit, can put `to: [{` all onto the same line. same for closing brakets.

src/api/sendgrid.ts
```diff
@@ -0,0 +1,53 @@
...
+                {
+                    email: 'UncleBu1728@gmail.com',
```
weihanwang, line 40: again, avoid checking in your real email to git.

src/api/sendgrid.ts
```diff
@@ -0,0 +1,53 @@
...
+        // edit accordingly on the template parameters
+        dynamic_template_data: {
+            receiver: 'J',
+            new_user_email: new_user_email,
+            sender: 'Yuanjian Admin Group',
+        },
+    }];
+
+    await sendEmail(personalizations, template_id);
+}
```
weihanwang, line 53: need an extra carriage return.

weihanwang (changes requested):

weihanwang (approved):

weihanwang:
merged as pr/33

PR #52, opened 2023-06-28
Yichi check concurrent meetings

UncleBu1728:
check meetings in progress per Tencent user-id every 2 min


src/api/TencentMeeting.ts
```diff
@@ -355,3 +355,39 @@ export async function getRecordURLs(meetingRecordId : string) {
 
 // Uncomment and modify this line to debug TM APIs.
```
weihanwang, line 356: leave these two lines at the bottom of file.

src/api/TencentMeeting.ts
```diff
@@ -355,3 +355,39 @@ export async function getRecordURLs(meetingRecordId : string) {
...
 // listRecords().then(res => console.log(JSON.stringify(res, null, 2)));
+
+export async function getCurrentMeetings() {
```
weihanwang, line 359: This file already has a function called listMeetings().

src/api/routes/cron.ts
```diff
@@ -39,6 +42,28 @@ const cron = router({
     });
     await Promise.all(promises);
   }),
+
+  /**
+   * Check the numbers of meetings in progess by each Tencent user id
```
weihanwang, line 47: The comment is confusing. We don't have multiple tencent users yet. so no need to mention "each". We should update the comment when multi-user support is added.

script.ts
weihanwang, line 1: What is this file for?

src/api/database/models/CurrentMeetings.ts
```diff
@@ -0,0 +1,21 @@
...
+  ForeignKey,
+  Table,
+  BelongsTo,
+  Model,
+  PrimaryKey
+} from "sequelize-typescript";
+import { INTEGER, STRING } from "sequelize";
+
+@Table({ paranoid: true })
+class CurrentMeetings extends Model {
```
weihanwang, line 12: `OngoingMeetingCount`. And use singular form.

src/api/routes/cron.ts
```diff
@@ -39,6 +42,28 @@ const cron = router({
...
+   */
+
+  currentMeetingNumbers: procedure.mutation(async () => {
+    const TMuserID: string = "ADMIN"; // change if multiple users used, possibly using hashed TM user Ids
```
weihanwang, line 51: `TMUserId`. Also, use TM_ADMIN_USER_ID and not a fake id.

src/api/routes/cron.ts
```diff
@@ -39,6 +42,28 @@ const cron = router({
...
+    const currentMeetingList = await getCurrentMeetings();
+    const currentMeetingNum = currentMeetingList.meeting_info_list.filter(obj => obj.status === 'MEETING_STATE_STARTED').length;
+
+    invariant(Number.isInteger(currentMeetingNum));
+
+    const userMeetings = await CurrentMeetings.findOrCreate({ where: { TMuserId: TMuserID } })
```
weihanwang, line 57: simply call `upsert()`

src/api/database/models/CurrentMeetings.ts
```diff
@@ -0,0 +1,21 @@
...
+  @PrimaryKey
+  @Column(STRING)
+  TMuserId: string;
```
weihanwang, line 15: `TMUserId`

src/api/database/models/CurrentMeetings.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  @Column(INTEGER)
+  currentMeetings: number | null;
```
weihanwang, line 18: `count`. Also, don't allow `null`.

src/api/routes/cron.ts
```diff
@@ -39,6 +42,28 @@ const cron = router({
     });
     await Promise.all(promises);
   }),
+
+  /**
+   * Check the numbers of meetings in progess by each Tencent user id
+   */
+
+  currentMeetingNumbers: procedure.mutation(async () => {
```
weihanwang, line 50: `updateOngoingMeetingCounts`

src/api/database/models/CurrentMeetings.ts
```diff
@@ -0,0 +1,21 @@
...
+  Column,
+  ForeignKey,
+  Table,
+  BelongsTo,
+  Model,
+  PrimaryKey
+} from "sequelize-typescript";
+import { INTEGER, STRING } from "sequelize";
+
+@Table({ paranoid: true })
```
weihanwang, line 11: Why do you need paranoid mode here?

src/api/routes/cron.ts
```diff
@@ -39,6 +42,28 @@ const cron = router({
...
+    const TMuserID: string = "ADMIN"; // change if multiple users used, possibly using hashed TM user Ids
+    const currentMeetingList = await getCurrentMeetings();
+    const currentMeetingNum = currentMeetingList.meeting_info_list.filter(obj => obj.status === 'MEETING_STATE_STARTED').length;
+
+    invariant(Number.isInteger(currentMeetingNum));
+
+    const userMeetings = await CurrentMeetings.findOrCreate({ where: { TMuserId: TMuserID } })
+    console.log(userMeetings);
```
weihanwang, line 58: don't check in unnecessary log lines.

src/api/routes/cron.ts
```diff
@@ -39,6 +42,28 @@ const cron = router({
...
+  /**
+   * Check the numbers of meetings in progess by each Tencent user id
+   */
+
+  currentMeetingNumbers: procedure.mutation(async () => {
+    const TMuserID: string = "ADMIN"; // change if multiple users used, possibly using hashed TM user Ids
+    const currentMeetingList = await getCurrentMeetings();
+    const currentMeetingNum = currentMeetingList.meeting_info_list.filter(obj => obj.status === 'MEETING_STATE_STARTED').length;
+
+    invariant(Number.isInteger(currentMeetingNum));
```
weihanwang, line 55: I don't think this assertion is useful. JS engine guarantees length always returns an integer.

src/api/routes/cron.ts
```diff
@@ -39,6 +42,28 @@ const cron = router({
...
     await Promise.all(promises);
   }),
+
+  /**
+   * Check the numbers of meetings in progess by each Tencent user id
+   */
+
+  currentMeetingNumbers: procedure.mutation(async () => {
+    const TMuserID: string = "ADMIN"; // change if multiple users used, possibly using hashed TM user Ids
+    const currentMeetingList = await getCurrentMeetings();
```
weihanwang, line 52: spare the intermediate variable. `const count = (await listMeetings()).meeting_info_list...`.

vercel.json
```diff
@@ -1,6 +1,12 @@
...
-  }]
-}
+  "crons": [
+    {
+      "path": "/api/v1/cron.uploadRawTranscripts",
+      "schedule": "30 9 * * *"
+    },
+    {
+      "path":"/api/v1/cron.currentMeetingNumbers",
+      "schedule": "2 * * * *"
```
weihanwang, line 9: fwiw, the free vercel version doesn't allow cron jobs more frequently than once a day. we can change it to a minute or two once we get vercel's open source discount.

vercel.json
```diff
@@ -1,6 +1,12 @@
...
+    }
+  ]
+}
```
weihanwang, line 12: missing new line at eof.

weihanwang (changes requested):

src/api/routes/cron.ts
UncleBu1728: Do we need to hash the id instead of inputting the actual id?

src/api/TencentMeeting.ts
```diff
@@ -353,5 +353,41 @@ export async function getRecordURLs(meetingRecordId : string) {
   return res;
 }
 
+export async function countOngoingMeeting() {
```
weihanwang, line 356: My point was why not just use the existing `listMeetings` function? There is no need to fine tune zRes just because we don't need some fields.

src/api/database/models/OngoingMeetingCount.ts
```diff
@@ -0,0 +1,21 @@
...
+  ForeignKey,
+  Table,
+  BelongsTo,
+  Model,
+  PrimaryKey
+} from "sequelize-typescript";
+import { INTEGER, STRING } from "sequelize";
+
+@Table
+class OngoingMeetingCount extends Model {
```
weihanwang, line 12: Write a comment to explain this model, for those who aren't familiar with the design. Something like "A map from TencentMeeting admin user ids to the number of ongoing meetings in the user's account. Normally we expect either 0 or 1 because...."

And talking about which, we should probably `TMAdminUserId` instead as the primary key.

src/api/routes/cron.ts
```diff
@@ -39,6 +42,20 @@ const cron = router({
     });
     await Promise.all(promises);
   }),
+
+  /**
+   * Check the numbers of meetings ongoing and update the OngoingMeetingCount table
```
weihanwang, line 47: `ongoing meetings`

weihanwang (changes requested):

vercel.json
weihanwang: Actually you need to change this:


Deployment failed with the following error:

`Hobby accounts are limited to daily cron jobs. This cron expression (2 * * * *) would run more than once per day. Upgrade to pro to unlock all Cron Jobs features on Vercel.`

src/api/routes/cron.ts
```diff
@@ -39,6 +42,20 @@ const cron = router({
     });
     await Promise.all(promises);
   }),
+
+  /**
+   * Check the ongoing meetings and update the OngoingMeetingCount table
+   */
+
```
weihanwang, line 49: remove empty line here

src/api/database/models/OngoingMeetingCount.ts
weihanwang: `TMAdminUserId`

weihanwang (changes requested):

PR #54, opened 2023-06-30
added counting and warning ongoing meetings

UncleBu1728:
added an api in myGroup to fetch the count of ongoing groups
added a modal to warn users when there are concurrent meetings and prevent them from entering meetings


src/api/routes/myGroups.ts
```diff
@@ -23,56 +25,64 @@ const myGroups = router({
    * TODO: Only allow group users to call this function.
    */
   generateMeetingLink: procedure
-  .use(authUser())
```
weihanwang, line 26: Can you undo these large-scale auto formatting changes? It makes review very difficult.

We can check in a PR dedicated to format changes.

Also, please self-review a PR before submitting for review. Thanks~

weihanwang (changes requested):

src/components/GroupBanner.tsx
```diff
@@ -27,20 +35,44 @@ export default function GroupBanner(props: {
...
-    } finally {
-      // More time is needed to redirect to the meeting page. Keep it spinning.
-      // We should uncomment this line if we pop the page in a new window.
-      // setJoining(false);
+    if ((await tClientBrowser.myGroups.countOngoingMeeting.query()) !== 0) {
+      setOccupancy(true);
+    } else {
+      setJoining(true);
+      try {
+        const link = await tClientBrowser.myGroups.generateMeetingLink.mutate({ groupId: groupId });
```
weihanwang: Also, we need to change the table design of `OngoingMeetingCount`, because we should allow expected participants of the ongoing meeting to join. We need to change it to:

```
class OngoingMeetings ... {
  @PrimaryKey
  TMAdminUserId: string;
  
  @PrimaryKey
  meetingLink: string;
}
```

`joinMeeting` should first see if the current, non-expired meeting link is already in this table, if yes return the link. if no, then check the row count of this table.

Can you create a new PR just for refactoring this table and `updateOngoingMeetingCounts`? Then we can come back to this PR.

weihanwang (changes requested):

PR #56, opened 2023-07-01
Set up viewport height of card box to achieve sticky footer instead of using fixed footer position

Linco1n3:


Linco1n3:
https://watch.screencastify.com/v/hFDR4LNcxJEaz6TVJ2sr

Linco1n3:
It still looks like a short-term solution. I was trying to set the container which holds navBar, main content(child cards), and footer as a flex box, but it's a bit more complicated than I thought. 
Btw I saw there is a task to refactor the AppLayout and get rid of horizon-ui. Happy to help with it if needed.

weihanwang:
Nice! Yes! if you are interested, please take that task of refactoring AppLayout and remove horizon-ui. Thank you!

weihanwang (approved):

PR #73, opened 2023-07-04
Adding index on user.roles to speed up roles-related queries.

aoranwu:
I add a GIN index on roles field of users table to speed up roles-related queries. Is this the right type of index to use for JSONB-typed field?

Confirmed in Neon: image


src/api/database/models/User.ts
```diff
@@ -37,6 +38,10 @@ class User extends ParanoidModel<
   @Column(STRING)
   clientId: string;
 
+  @Index({
+    name: 'user-roles-index',
```
weihanwang, line 42: Do you really need to specify the name?

weihanwang:
@aoranwu Did you verify the working of the index by using `sql explain` or alike? In the code we use sequelize's Op.contains to search into the column:

https://github.com/yuanjian-org/app/blob/3c1998396f5975d2a0e9e50c9f3607242298e59b/src/api/sendgrid.ts#L83

https://github.com/yuanjian-org/app/blob/3c1998396f5975d2a0e9e50c9f3607242298e59b/src/api/routes/users.ts#L129

aoranwu:
> @aoranwu Did you verify the working of the index by using `sql explain` or alike? In the code we use sequelize's Op.contains to search into the column:
> 
> https://github.com/yuanjian-org/app/blob/3c1998396f5975d2a0e9e50c9f3607242298e59b/src/api/sendgrid.ts#L83
> 
> https://github.com/yuanjian-org/app/blob/3c1998396f5975d2a0e9e50c9f3607242298e59b/src/api/routes/users.ts#L129

The indexed is used when I inserted ~5000 fake data entries: 
image

When there're fewer records in the database (~1000, for example), the query performs sequential scan without using the index.

src/api/database/models/User.ts
aoranwu: 'user.roles' will be auto-generated to be the name of the index so removed.

weihanwang (approved):

PR #74, opened 2023-07-05
Remove dropping db option from `yarn sync-database`

weihanwang:
If users really want to start over, the cleanest way is to drop db and recreate a new one.


lt616:
Thanks for making the change!

addlistener (approved):

PR #76, opened 2023-07-07
hot fix on editable usage

lt616:
- onChange should be applied to <Editable> tag not the child <input> tag, otherwise the editable might not working well
- remove unused import


lt616:
Found an issue that the Editable should not be refreshed when content change, otherwise it will change from editing mode to previewing mode when user typing new name. Added useMemo block so that Editable won't refresh when name changes.

PR #78, opened 2023-07-08
finished getSpeakerTime.ts

FangkuaikuaiOffical:


src/shared/getSpeakerTime.ts
weihanwang, line 1: 1. I expect only backend would use it. so can move to `src/api`.
2. how about `parseSpeakerStats` instead? We may add more stats in addition to speaking time later on.
3. Can you add some test cases especially to cover corner cases such as empty string, a single line of transcript, two lines of transcript, incomplete transcript, malformed transcript.

src/shared/getSpeakerTime.ts
```diff
@@ -0,0 +1,51 @@
+import fs from 'fs';
+
+interface Speaker {
```
weihanwang, line 3: `SpeakerStats` instead

src/shared/getSpeakerTime.ts
```diff
@@ -0,0 +1,51 @@
...
+  name: string;
+  totalSpeakingSeconds: number;
+}
+
+export function getSpeakerTime(filePath: string): Promise<Speaker[]> {
```
weihanwang, line 8: 1. pass in the transcript string directly instead of a file. we aren't going to store transcripts as files.
2. without file reading it won't need to return Promise any more.

PR #112, opened 2023-07-25
add import rule and fix to use relative paths

UncleBu1728:
added `no-restricted-imports` rule to prevent users from importing project files (/src/*) as custom modules.
- Note that the paths and patterns of this rule are manually entered, so if any files or directories under src/ changed, they might also need to be updated in the rule

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,6 @@
 {
   "extends": "next/core-web-vitals",
+  "plugins": ["import"],
   "rules": {
     "object-curly-spacing": ["warn", "always"],
     "import/no-restricted-paths": ["error",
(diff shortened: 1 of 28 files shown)


.eslintrc.json
```diff
@@ -23,6 +24,31 @@
         "message": "Files outside of `src/api` must not refer to files in `src/api`. See README.md"
       }]
     }],
-    "import/no-absolute-path":"error"
+    "import/no-absolute-path":"error",
+    "no-restricted-imports": [
+      "error",{
+        "paths":[
+          // files under src
+        "AppLayout",
```
weihanwang, line 32: why do we need to specify files? I don't think it's a workable solution as we will constantly add and rename files.

PR #113, opened 2023-07-27
Delete user zlu

FangkuaikuaiOffical:
cascade delete on groups is done.
The database prevents deletion of a user if there is a reference to the user in the partnership, even though the partnership already has the deletedAt row populated. This might be a bug.
Could there be a permission to cascade delete partnerships when deleting users? This is the opposite of the current database setup.


weihanwang:
What is the error message you observed when trying to cascade delete partnership?

FangkuaikuaiOffical:
> What is the error message you observed when trying to cascade delete partnership?

截屏2023-08-09 02 31 08
this is what looks like when I try delete a user directly in pgAdmin

```
1) User
       cascadeDestory
         should create test users:
     Validation error
  Error
      at Query.run (node_modules/sequelize/src/dialects/postgres/query.js:76:25)
      at /Users/ziyi/Desktop/Visionary Education/app/node_modules/sequelize/src/sequelize.js:650:28
      at async PostgresQueryInterface.insert (node_modules/sequelize/src/dialects/abstract/query-interface.js:795:21)
      at async User.save (node_modules/sequelize/src/model.js:4153:35)
      at async Function.create (node_modules/sequelize/src/model.js:2305:12)

  2) User
       cascadeDestory
         should create a group related to the users:
     TypeError: Cannot read properties of undefined (reading 'id')
      at /Users/ziyi/Desktop/Visionary Education/app/src/api/database/models/User.test.ts:80:51
      at Generator.next (<anonymous>)
      at fulfilled (src/api/database/models/User.test.ts:5:58)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  3) User
       cascadeDestory
         should create a partnership between test users:
     TypeError: Cannot read properties of undefined (reading 'id')
      at /Users/ziyi/Desktop/Visionary Education/app/src/api/database/models/User.test.ts:88:25
      at Generator.next (<anonymous>)
      at /Users/ziyi/Desktop/Visionary Education/app/src/api/database/models/User.test.ts:8:71
      at new Promise (<anonymous>)
      at __awaiter (src/api/database/models/User.test.ts:4:12)
      at Context.<anonymous> (src/api/database/models/User.test.ts:86:69)
      at processImmediate (node:internal/timers:478:21)

  4) User
       cascadeDestory
         should delete the partnership and verify deletion:
     TypeError: Cannot read properties of undefined (reading 'destroy')
      at /Users/ziyi/Desktop/Visionary Education/app/src/api/database/models/User.test.ts:112:25
      at Generator.next (<anonymous>)
      at /Users/ziyi/Desktop/Visionary Education/app/src/api/database/models/User.test.ts:8:71
      at new Promise (<anonymous>)
      at __awaiter (src/api/database/models/User.test.ts:4:12)
      at Context.<anonymous> (src/api/database/models/User.test.ts:110:72)
      at processImmediate (node:internal/timers:478:21)

  5) User
       cascadeDestory
         should delete user2 and all related GroupUser instances:
     TypeError: Cannot read properties of undefined (reading 'destroy')
      at /Users/ziyi/Desktop/Visionary Education/app/src/api/database/models/User.test.ts:141:19
      at Generator.next (<anonymous>)
      at /Users/ziyi/Desktop/Visionary Education/app/src/api/database/models/User.test.ts:8:71
      at new Promise (<anonymous>)
      at __awaiter (src/api/database/models/User.test.ts:4:12)
      at Context.<anonymous> (src/api/database/models/User.test.ts:140:78)
      at processImmediate (node:internal/timers:478:21)

  6) User
       cascadeDestory
         should delete test user1 and test group:
     update or delete on table "users" violates foreign key constraint "Partnerships_mentorId_fkey" on table "Partnerships"
  Error
      at Query.run (node_modules/sequelize/src/dialects/postgres/query.js:76:25)
      at /Users/ziyi/Desktop/Visionary Education/app/node_modules/sequelize/src/sequelize.js:650:28
      at async PostgresQueryInterface.bulkDelete (node_modules/sequelize/src/dialects/abstract/query-interface.js:992:12)
      at async Function.destroy (node_modules/sequelize/src/model.js:3082:16)
```

this is the error message when the test file DO NOT use force mode to delete partership.

It seems that when the database deletes, it does not judge whether there is a valid association based on the deletedAt value, but judges whether the row exists in the database. That is to say, the database did not correctly apply the paranoid mode when judging.
Another phenomenon that proves this guess is that there is also an error in the database that does not allow adding "users with the same email address that has been soft-deleted".

weihanwang:
* The error from pgAdmin is WAI (working as intended) because you need to manually delete the partnership before the user row can be physically deleted.
* The error "TypeError: Cannot read properties of undefined (reading 'id')" seems unrelated but may cause later steps to fail. can you fix those errors first?
* Not allowing inserting a soft-deleted user is also WAI. we will need to use `upsert` to add new users.
* "It seems that when the database deletes, it ..." That is correct. sequelize doesn't detect or support cascaded paranoid mode. that why we need to manually cascade delete in the first place.

FangkuaikuaiOffical:
Modifications have been made as required. Due to the problem of the Role component, the yarn test under main cannot run normally for the time being. The previous tests all passed.

src/api/database/models/User.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        }
+      });
+
+      // 确认找到的partnership存在
+      expect(foundPartnership).to.exist;
+    });
+
+
+
+
```
weihanwang, line 106: remove extra empty lines

src/api/database/models/User.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    // 删除行为开始
+    // 检测partnership中的deletedAt是否存在值
+
+    // ...
```
weihanwang, line 110: remove unuseful comments

src/api/database/models/User.ts
```diff
@@ -5,6 +5,8 @@ import type {
 } from "sequelize";
 import {
   AllowNull,
+  BeforeBulkDestroy,
```
weihanwang, line 8: it doesn't seem to be referred to anywhere?

src/api/database/models/User.ts
```diff
@@ -69,6 +74,30 @@ class User extends ParanoidModel<
...
+
+
+
+  @BeforeDestroy
+  static async cascadeSoftDelete(user: User, options: any) {
+
+    const moment = require('moment-timezone');
+    const timestamp = moment().format('YYYY-MM-DD HH:mm:ss.SSSZ');
+
+    await GroupUser.update(
```
weihanwang, line 87: I don't think it's the right thing to do. @BeforeDestroy needs to support both soft and hard deletion.

weihanwang (changes requested):

FangkuaikuaiOffical:
updated @cascadeDestory to fit both soft and hard delete mode.
updated test case to fit new database changes.
all tests passed.

src/api/database/models/User.ts
```diff
@@ -69,6 +73,41 @@ class User extends ParanoidModel<
 
   @HasMany(() => Interview)
   interviews: NonAttribute<Interview[]>;
+
+  @BeforeDestroy
+  static async cascadeDelete(user: User, options: any) {
+
+    if (options.force) {
```
weihanwang, line 80: why do you need to distinguish forced and non-forced deletion? can you just do:

```
await GroupUser.destroy({
        where: { userId: user.id },
        ...options,
      });
```

in both cases? same for partnership.

src/api/routes/users.ts
```diff
@@ -247,6 +247,29 @@ const listPriviledgedUserDataAccess = procedure
...
 });
 
+const remove = procedure
+  .use(authUser())
+  .input(z.object({
+    id: z.string(),
+  }))
+  .mutation(async ({ input, ctx }) => 
+{
+  const isUserOrPRManager = isPermitted(ctx.user.roles, ['UserManager', 'PrivilegedRoleManager']);
```
weihanwang, line 257: 1. only UserManager is necessary.
2. use `authUser()` to simplify the logic. so:

```
  .user(authUser("UserManager"))
```

and remove this text.

src/api/routes/users.ts
```diff
@@ -247,6 +247,29 @@ const listPriviledgedUserDataAccess = procedure
   });
 });
 
+const remove = procedure
```
weihanwang, line 250: "remove" => "delete" to be consistent with sql terms

src/pages/users.tsx
```diff
@@ -173,19 +181,22 @@ function UserEditor(props: {
...
+                    </Checkbox>
+                  );
+                })}
             </Stack>
           </FormControl>
         </VStack>
       </ModalBody>
       <ModalFooter>
-        <Button variant='brand' isLoading={saving} onClick={save} isDisabled={!validEmail || !validName}>保存</Button>
+        <Flex justifyContent="space-between" width="100%">
```
weihanwang, line 196: I think `width="100%"` is not needed?

src/api/routes/users.ts
FangkuaikuaiOffical: could not use word "delete" because it is a reserved word in Javascript. How about "deleteUser" ?

src/pages/users.tsx
FangkuaikuaiOffical: Without this two buttons would be both in the lower right corner

src/api/routes/users.ts
weihanwang: ah. okay then remove is good

weihanwang (approved):

PR #121, opened 2023-07-31
Update tsconfig.json settings

UncleBu1728:
- `*cy.ts files` is excluded in `tsconfig.json` in the root file to prevent  `--isolatedModules` flag when running script `next build`
- a `tsconfig.json` file is added to the cypress folder to include type definitions from cypress, otherwise `*.cy.ts` files will show errors `'cy is not found'`


tsconfig.json
```diff
@@ -36,10 +36,10 @@
     },
     "incremental": true
   },
-  "exclude": ["node_modules"],
+  "exclude": ["node_modules","*/**/*.cy.ts"],
```
weihanwang, line 39: can you make a comment that cy.ts are covered in tsconfig.json?

tsconfig.json
```diff
@@ -36,10 +36,10 @@
...
   "include": [
     "./src",
     "src/api",
     "src/shared",
-    "next-env.d.ts", "**/*.ts", "**/*.tsx", "next.config.js"],
+    "next-env.d.ts", "**/*.ts", "**/*.tsx", "next.config.js", "cypress/tsconfig.json"],
```
weihanwang, line 44: 1. please use one dedicated line for each item
2. are you sure including ts.config.json is the right thing to do?

cypress/tsconfig.json
weihanwang, line 1: please also rename login.test.ts to login.cy.ts

cypress/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+    "compilerOptions": {
+      "target": "es5",
+      "lib": ["es5", "dom"],
+      "types": ["cypress", "node"]
+    },
+    "include": ["**/*.ts"]
+}
```
weihanwang, line 8: missing new line at eof

weihanwang (changes requested):

tsconfig.json
```diff
@@ -36,10 +36,17 @@
...
+    "*/**/*.cy.ts" /* .cy.test files are covered by cypress/tsconfig */
+  ],
   "include": [
     "./src",
     "src/api",
     "src/shared",
-    "next-env.d.ts", "**/*.ts", "**/*.tsx", "next.config.js"],
+    "next-env.d.ts",
+    "**/*.ts",
+    "**/*.tsx",
```
weihanwang, line 49: given these two lines, do you still need lin 44, 45, 46, 47, 50? can you remove them?

also, if to keep line 44, change it to `src`.

tsconfig.json
```diff
@@ -41,12 +41,8 @@
     "*/**/*.cy.ts" /* .cy.test files are covered by cypress/tsconfig */
   ],
   "include": [
-    "./src",
-    "src/api",
-    "src/shared",
-    "next-env.d.ts",
+    "src",
```
weihanwang, line 44: do you need this line?

weihanwang (approved):

PR #122, opened 2023-08-01
fix user profile simple grid width

lt616:
Replace hardcoded width with simple grid width to fix the display issue on mobile side


lt616:
Hi yaohan, please let me know if this has fixed the mobile side issue. thank you

weihanwang (approved):

PR #129, opened 2023-08-04
add summarization script

yangyingxiang:


llm/summarize.py
```diff
@@ -0,0 +1,233 @@
...
+url_get = "https://beta.yuanjian.org/api/v1/summaries.list"
+url_post = 'https://beta.yuanjian.org/api/v1/summaries.write'
+params = {"key": "原始文字", "excludeTranscriptsWithKey": "xxx_llm_1000"}
+headers_get = {"Authorization": "Bearer {}".format(os.environ['YUANJIAN_TOKEN'])}
+headers_post = ["Authorization: Bearer {}".format(os.environ['YUANJIAN_TOKEN']),
+                "Content-Type: application/x-www-form-urlencoded"]  #
+
+
+def get_list(url, headers, params):
+    """
```
weihanwang, line 41: Can we format to two spaces per tab?

llm/summarize.py
```diff
@@ -0,0 +1,233 @@
+"""
+Please set environment variable YUANJIAN_TOKEN for beta.yuanjian.org
```
weihanwang, line 2: Change to INTEGRATION_AUTH_TOKEN to be consistent with the rest of  codebase.

llm/summarize.py
yangyingxiang: PEP 8 recommends using 4 spaces, can we keep using that in python code?

weihanwang: sure thing. (was mentioning it because all js code in this repo use 2 spaces).

weihanwang (approved):

PR #130, opened 2023-08-07
Add Docker configurations.

lijunray:
Created Docker-related configurations and instructions.

Details:
  - Added a new dependency `sharp`  in `package.json` as recommended by [NextJS](https://nextjs.org/docs/messages/install-sharp).
  - Forked NextJS official [`Dockerfile`](https://github.com/vercel/next.js/blob/canary/examples/with-docker-compose/next-app/prod.Dockerfile), but it is a bit outdated so made some changes based on it.

More potential optimizations:
  - https://github.com/vercel/next.js/discussions/16995.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "private": true,
   "dependencies": {
+    "sharp": "0.32.4",
     "@authing/guard-react18": "^5.1.11",
     "@chakra-ui/icons": "^1.1.5",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "private": true,
   "dependencies": {
+    "sharp": "0.32.4",
```
weihanwang, line 6: what is this library for?

weihanwang (approved):

PR #136, opened 2023-08-13
Add Sentry for error tracking and monitoring

UncleBu1728:
Add and install Sentry and tweak configuration for production
To log in and see error reporting, check `Admin 账户` page on notion (originally Tencent meeting page)


.gitignore
```diff
@@ -37,3 +37,6 @@ yarn-error.log*
 *.tsbuildinfo
 
 .idea
+
+# Sentry Auth Token
+.sentryclirc
```
weihanwang, line 42: can't the auth token be placed in .env files? also, if we can't use .env, how to supply this auth token to vercel?

next.config.js
```diff
@@ -6,3 +6,41 @@ const nextConfig = {
 };
 
 module.exports = nextConfig;
+
+
+// Injected content via Sentry wizard below
```
weihanwang, line 11: can you comment on where this block of code was copied from?

package.json
```diff
@@ -111,4 +112,4 @@
     "regenerator-runtime": "^0.13.11",
     "ts-node": "^10.9.1"
   }
-}
+}
```
weihanwang, line 115: missing cr in eof

sentry.client.config.ts
weihanwang, line 1: do these files have to be in the root folder? can they be under /sentry?

sentry.client.config.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  // Adjust this value in production, or use tracesSampler for greater control
+  tracesSampleRate: 1,
+
+  // Setting this option to true will print useful information to the console while you're setting up Sentry.
+  debug: false,
+
+  environment: "production",
+
+  tracePropagationTargets: [/^https:\/\/beta\.yuanjian\.org/],
```
weihanwang, line 20: can you use /.*\.yuantu\..*/ and /.*\.yuanjian\..*/? what's the downside if we set the target too widely, e.g. /.*/ ?

next.config.js
```diff
@@ -6,3 +6,41 @@ const nextConfig = {
...
+  module.exports,
+  {
+    // For all available options, see:
+    // https://github.com/getsentry/sentry-webpack-plugin#options
+
+    // Suppresses source map uploading logs during build
+    silent: true,
+
+    org: "visionary-education-foundation",
+    project: "javascript-nextjs",
```
weihanwang, line 25: "app"

next.config.js
UncleBu1728: this is auto-generated when install and init sentry

sentry.client.config.ts
UncleBu1728: The default value was `['localhost', /^\//]`, I was thinking to limit specifically to `beta.yuanjian.org` to filter unnecessary errors that happened during development. `/.*/` will be used for the latest update

sentry.client.config.ts
UncleBu1728: I am searching for a solution to this but it seems there is no official support to change this easily. https://github.com/getsentry/sentry-javascript/issues/4249
I would need to check on some methods mentioned in this issue

.gitignore
UncleBu1728: This could be done using .env variables, this was automatically set up by the sentry default installation tool. https://docs.sentry.io/product/cli/configuration/#configuration-values

next.config.js
```diff
@@ -1,8 +1,52 @@
...
+// with Sentry.
+// https://nextjs.org/docs/api-reference/next.config.js/introduction
+
+const { withSentryConfig } = require("@sentry/nextjs");
+
+// your existing module.exports or default export
 const nextConfig = {
-  reactStrictMode: true,
-  swcMinify: true,
-  output: 'standalone',
```
weihanwang, line 5: Why are these configs removed?

weihanwang (approved):

weihanwang (approved):

PR #141, opened 2023-08-16
Tweak husky hooks yz

UncleBu1728:
Git `pre-push` hook is used instead of `pre-commit` to avoid running `yarn build` on each local commit
Mac users will require to run `chmod ug+x .husky/*` in order for git hooks to work, the instruction is added to README.md
- need to discuss whether to add this command to the script


.husky/pre-push
```diff
@@ -2,3 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 yarn lint
+yarn build
```
weihanwang, line 5: I think we just need "yarn test". vercel will do lint and build automatically.

package.json
```diff
@@ -75,8 +75,8 @@
     "nibble": "eslint-nibble --ext .jsx,.js,.tsx,.ts src",
     "cypress-run": "cypress run",
     "cypress-open": "cypress open",
-    "prepare": "if [ \"$VERCEL_ENV\" = \"production\" ]; then ts-node tools/syncDatabase.ts; fi",
-    "docker-dev": "docker compose up --build -d"
+    "docker-dev": "docker compose up --build -d",
+    "prepare": "husky install && if [ \"$VERCEL_ENV\" = \"production\" ]; then ts-node tools/syncDatabase.ts; fi"
```
weihanwang, line 79: why install hasky here?

package.json
```diff
@@ -105,7 +105,7 @@
     "eslint-config-next": "^12.2.5",
     "eslint-nibble": "^8.1.0",
     "gh-pages": "^3.2.3",
-    "husky": "^8.0.3",
+    "husky": "^8.0.0",
```
weihanwang, line 108: why downgrading?

.husky/pre-push
weihanwang: however, yarn test should only be run locally and not in staging (aka development) or production environments, because it modifies the database.

.husky/common,sh
weihanwang, line 1: common,sh or common.sh?

package.md
weihanwang, line 1: package.json.md

package.md
```diff
@@ -0,0 +1,6 @@
+# notes on package.json
+
+## scripts
+ `"prepare"`
+- `husky install` is required by husky official doc, see details: https://typicode.github.io/husky/getting-started.html
+-  `if [ \"$VERCEL_ENV\" = \"production\" ]` statement is to check if the application is deployed on Vercel, then run `yarn sync-database` on each deployment. yarn
```
weihanwang, line 6: is the last word "yarn" a typo?

.husky/pre-push
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn lint
+# yarn sync-database is required before testing to prevent errors
```
weihanwang, line 4: why doesn't this script automatically run yarn sync-database then?

package.json.md
```diff
@@ -0,0 +1,6 @@
+# notes on package.json
+
+## scripts
+ `"prepare"`
```
weihanwang, line 4: Can you describe when is prepare invoked?

weihanwang (approved):

PR #145, opened 2023-08-18
Optimize summary list query

UncleBu1728:
Merged two queries and filter into one query to return desired data

### DIFF
--- a/src/api/routes/summaries.ts
+++ b/src/api/routes/summaries.ts
@@ -9,6 +9,8 @@ import Group from "../database/models/Group";
 import { TRPCError } from "@trpc/server";
 import { safeDecodeMeetingSubject } from "./meetings";
 import apiEnv from "api/apiEnv";
+import { Sequelize, Op } from "sequelize";
+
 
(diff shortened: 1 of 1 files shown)


src/api/routes/summaries.ts
```diff
@@ -47,23 +49,21 @@ const list = procedure
...
-  const summaries = await Summary.findAll({ 
-    where: { summaryKey: input.key },
-    attributes: ['transcriptId', 'summary'],
+  ).query(async ({ input }: { input: SummariesListInput }) => {
+    return await Summary.findAll({
+      where: {
+        summaryKey: input.key,
+        transcriptId: {
+          [Op.notIn]: Sequelize.literal(`
+            (SELECT "transcriptId" FROM "Summaries"
```
weihanwang, line 58: I'd much prefer not to use raw select queries for performance reasons. these queries are very hard to maintain and easy to introduce bugs.

Actually, because we need to have a remodel the summary upload API and it's very likely that this function will need a complete rewrite, let's not worry about this optimization task for now. Sorry about the late notice.

PR #160, opened 2023-08-29
Update Sentry config and environment variables

UncleBu1728:
updated `next.config.js`
`environment` is adaptively changed to either "production" or "development"
`project` name is now changed to "app"
`SENTRY_AUTH_TOKEN` in `.env` is removed
`hideSourceMaps` is now `false`, sentry issues should show where the error happened in the source code. Further tests needed 
`ignoreErrors` is added to ignore any issues/errors from "localhost“

passed on developing both on local and Vercel

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -24,7 +24,7 @@ module.exports = withSentryConfig(
     silent: true,
 
     org: "visionary-education-foundation",
-    project: "javascript-nextjs",
+    project: "app",
   },
   {
     // For all available options, see:
--- a/next.config.js
+++ b/next.config.js
@@ -40,7 +40,7 @@ module.exports = withSentryConfig(
     tunnelRoute: "/monitoring",
 
     // Hides source maps from generated client bundles
-    hideSourceMaps: true,
+    hideSourceMaps: false,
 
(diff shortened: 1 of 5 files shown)


sentry.client.config.ts
```diff
@@ -15,7 +15,12 @@ Sentry.init({
   // Setting this option to true will print useful information to the console while you're setting up Sentry.
   debug: false,
 
-  environment: "production",
+  environment: process.env.NODE_ENV,
 
   tracePropagationTargets: ['localhost', /^\//],
```
weihanwang, line 20: if localhost is ignored, do we still need localhost here?

next.config.js
```diff
@@ -40,7 +40,7 @@ module.exports = withSentryConfig(
     tunnelRoute: "/monitoring",
 
     // Hides source maps from generated client bundles
-    hideSourceMaps: true,
+    hideSourceMaps: false,
```
weihanwang, line 43: on line 37, how much bundle size turning off `transpileClientSDK` can save?

weihanwang (approved):

PR #168, opened 2023-09-05
Add Logging service using Sentry captureConsole

UncleBu1728:
`captureConcole` integration from Sentry and a customized function is added to log user information 
- console.error and console.info is now monitored by Sentry for logging purposes
- The implementation on` login.tsx `page is to monitor if the authing tokens are still valid when users login. Normally if a user logout should clean the local storage which should not contain keys of `_authing_token` nor `_authing_user`. Therefore, if the keys mentioned above exist when the user is directed to the login page, it could mean the authentication token is expired or invalid. The `logToSentry` function should send logging messages if such situations happen. 

con: captureConsole integration could capture any console logger errors from localhosts, but it could be filtered or removed by the searching options on Sentry/issue page


src/shared/logToSentry.ts
```diff
@@ -0,0 +1,26 @@
+import { z } from 'zod';
+// other details like timestamp, and hostname/ IP address should be included by Sentry
+const zLogObject = z.object({
+    message: z.string(),
+    email: z.string().email().optional(),
+    data: z.object({}).optional()
+})
```
weihanwang, line 7: Add semi-colon. Also, in a different PR, can you help enable eslint check for semi-colons?

src/shared/logToSentry.ts
```diff
@@ -0,0 +1,26 @@
+import { z } from 'zod';
+// other details like timestamp, and hostname/ IP address should be included by Sentry
```
weihanwang, line 2: add empty line

src/shared/logToSentry.ts
```diff
@@ -0,0 +1,26 @@
...
+const zLogObject = z.object({
+    message: z.string(),
+    email: z.string().email().optional(),
+    data: z.object({}).optional()
+})
+
+enum logLevel{
+    ERROR,
+    INFO
+}
```
weihanwang, line 12: semi-colon

src/shared/logToSentry.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const logToSentry = (tag: keyof typeof logLevel, logDetail: logObject) => {
+    if ((tag) === 'ERROR') {
+        console.error('LOG', tag, JSON.stringify(logDetail));
+    }
+
+    if ((tag) === 'INFO') {
+        console.info('LOG', tag, JSON.stringify(logDetail));
+    }
+}
```
weihanwang, line 24: semi-colon

src/shared/logToSentry.ts
```diff
@@ -0,0 +1,26 @@
+import { z } from 'zod';
+// other details like timestamp, and hostname/ IP address should be included by Sentry
+const zLogObject = z.object({
+    message: z.string(),
+    email: z.string().email().optional(),
+    data: z.object({}).optional()
+})
+
+enum logLevel{
```
weihanwang, line 9: `LogLevel`

src/shared/logToSentry.ts
```diff
@@ -0,0 +1,26 @@
...
+    ERROR,
+    INFO
+}
+
+export type logObject = z.TypeOf<typeof zLogObject>;
+
+const logToSentry = (tag: keyof typeof logLevel, logDetail: logObject) => {
+    if ((tag) === 'ERROR') {
+        console.error('LOG', tag, JSON.stringify(logDetail));
```
weihanwang, line 18: I don't understand why you need the logToSentry function. Can't we just use console.log instead whereever logToSentry is called?

src/shared/logToSentry.ts
```diff
@@ -0,0 +1,26 @@
+import { z } from 'zod';
+// other details like timestamp, and hostname/ IP address should be included by Sentry
+const zLogObject = z.object({
```
weihanwang, line 3: No need to use zod. define typescript types directly:

```
type LogObject = {
   ...
}
```

src/shared/logToSentry.ts
```diff
@@ -0,0 +1,26 @@
...
+})
+
+enum logLevel{
+    ERROR,
+    INFO
+}
+
+export type logObject = z.TypeOf<typeof zLogObject>;
+
+const logToSentry = (tag: keyof typeof logLevel, logDetail: logObject) => {
```
weihanwang, line 16: Why do you need the `LogObject` type? Why don't  just pass message, email, data as three function parameters?

src/pages/callback.tsx
weihanwang, line 1: Don't mix unrelated changes in one PR for reviews.

src/shared/logToSentry.ts
UncleBu1728: I was thinking of creating this object type just in case if we need more keys or properties to be passed to log without changing the number of parameters of `logToSentry`, but I did a quick research about the inefficiency of JSON.stringfy and will change it to a list parameters

src/shared/logToSentry.ts
UncleBu1728: In addition to what I have sent on WeChat, I was also thinking of keeping the logs in a consistent format and easier to manage

src/pages/login.tsx
```diff
@@ -34,10 +34,10 @@ export default function Login() {
...
         <Container as="footer">
           <Link isExternal href="https://vercel.com/?utm_source=yuanjian&utm_campaign=oss">
-            <Image 
-            src={vercelBanner} 
-            alt="Vercel Banner"  
-            height="30"
+            <Image
+              src={vercelBanner}
+              alt="Vercel Banner"
+              height="30"
```
weihanwang, line 40: can you revert this file? I'm removing it entirely.

sentry.client.config.ts
```diff
@@ -3,19 +3,23 @@
...
 import * as Sentry from "@sentry/nextjs";
+import { CaptureConsole } from '@sentry/integrations';
 
 Sentry.init({
   dsn: "https://a1dc8cd72a62aacab0dd520eee207398@o4505687617175552.ingest.sentry.io/4505687622090752",
 
-  integrations: [new Sentry.BrowserTracing()],
+  integrations: [
+    new Sentry.BrowserTracing(),
+    new CaptureConsole({ levels: ['error', 'info'] }) // using error and info as customized tag logging service
```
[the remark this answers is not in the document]
weihanwang: This is just a hypothesis, right? did you observe excessive messages already?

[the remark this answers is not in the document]
UncleBu1728: The reason I changed the environment to 'testing' earlier was that I observed there were logger messages sent to Sentry even with the `ignoreErrors` from 'localhost'. So I was thinking of using the logger levels that are not used very often to log messages like "info" or "error". If you go to the sentry/issue page and check the console logger message in development, you will see that they are sent from localhost

weihanwang (approved):

weihanwang (approved):

PR #182, opened 2023-09-19
Extend next-auth session data to include user fields

UncleBu1728:
- User data is included in the next-auth session and should be available by calling `useSession` instead of a `trpc.user.me query`.
- Type of `session.user` is declared as the `User` model.
- The website functions as normal including user login and new user signup.
- TODO: optimize redundant APIs and db calls with session data, such as removing trpc.user.me.query, and other unnecessary codes. Also check TODOs in auth.ts

### DIFF
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -26,19 +26,11 @@ export const authIntegration = () => middleware(async ({ ctx, next }) => {
 export const authUser = (permitted?: Role | Role[]) => middleware(async ({ ctx, next }) => {
   const session = await getServerSession(ctx.req, ctx.res, authOptions);
   if (!session) throw unauthorizedError();
-  const email = session.user?.email;
-  invariant(email);
 
-  // TODO: extend next-auth session data to include other user fields, and remove this extra call to db
-  const user = await db.User.findOne({
-    where: { email },
-    attributes: userAttributes,
-  });
-  invariant(user);
+  const roles = session.user?.roles;
+  if (!roles || !isPermitted(roles, permitted)) throw forbiddenError();
 
-  if (!isPermitted(user.roles, permitted)) throw forbiddenError();
-
-  return await next({ ctx: { user, baseUrl: ctx.baseUrl } });
+  return await next({ ctx: { user: session.user, baseUrl: ctx.baseUrl } });
 });
 
 const unauthorizedError = () => new TRPCError({
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -28,7 +28,8 @@ export const authUser = (permitted?: Role | Role[]) => middleware(async ({ ctx, 
   if (!session) throw unauthorizedError();
 
   const roles = session.user?.roles;
-  if (!roles || !isPermitted(roles, permitted)) throw forbiddenError();
+  invariant(roles);
+  if (!isPermitted(roles, permitted)) throw forbiddenError();
 
   return await next({ ctx: { user: session.user, baseUrl: ctx.baseUrl } });
 });
--- a/src/components/AppPageContainer.tsx
+++ b/src/components/AppPageContainer.tsx
@@ -6,30 +6,34 @@ import UserContext from "../UserContext";
 import User from '../shared/User';
 import NavBars, { sidebarBreakpoint, sidebarContentMarginTop, topbarHeight } from 'components/Navbars';
 import PageLoader from 'components/PageLoader';
-import { useSession } from 'next-auth/react';
+import { getSession } from 'next-auth/react';
 
 export default function AppPageContainer({ children, wide, ...rest }: {
   wide: boolean
 } & PropsWithChildren) {
   const [user, setUser] = useState<User | null>(null);
-  const { data: session } = useSession();
 
   useEffect(() => {
-    if (session) {
-      setUser(session.user);
-    }
-  }, [session]);
+    const fetchSession = async () => {
+      const session = await getSession();
+      if (session) {
+        setUser(session.user);
+      }
+    };
+
+    fetchSession();
+  }, []);
 
   return !user ?
     <PageLoader {...rest} />
-    : 
+    :
     <UserContext.Provider value={[user, setUser]}>
       <NavBars {...rest}>
         <Box
           marginTop={sidebarContentMarginTop}
-          paddingX={{ 
+          paddingX={{
             base: "16px",
-            [sidebarBreakpoint]: "30px" 
+            [sidebarBreakpoint]: "30px"
           }}
           maxWidth={{
             base: "100%",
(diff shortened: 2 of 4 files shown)


src/pages/api/auth/[...nextauth].ts
```diff
@@ -6,6 +6,15 @@ import { SendVerificationRequestParams } from "next-auth/providers";
 import { email as sendEmail, emailRoleIgnoreError } from "../../../api/sendgrid";
 import randomNumber from "random-number-csprng";
 import { toChinese } from "shared/strings";
+import { userAttributes } from "api/database/models/attributesAndIncludes";
+import invariant from "tiny-invariant";
+import User from "api/database/models/User";
```
Comment (weihanwang) [inline src/pages/api/auth/[...nextauth].ts:11]: remove this.

src/components/AppPageContainer.tsx
```diff
@@ -3,21 +3,22 @@ import Footer, { footerBreakpoint, footerMarginTop } from 'components/Footer';
...
 import User from '../shared/User';
 import NavBars, { sidebarBreakpoint, sidebarContentMarginTop, topbarHeight } from 'components/Navbars';
 import PageLoader from 'components/PageLoader';
+import { useSession } from 'next-auth/react';
 
 export default function AppPageContainer({ children, wide, ...rest }: {
   wide: boolean
 } & PropsWithChildren) {
   const [user, setUser] = useState<User | null>(null);
+  const { data: session } = useSession();
```
weihanwang, line 15: don't call useSession() here. pass in the session object from _app.tsx:SwitchBoard()

src/components/AppPageContainer.tsx
weihanwang, line 1: The TODO in auth.ts also needs to be addressed.

src/components/AppPageContainer.tsx
UncleBu1728: The session has been passed down properly with the provider already, so a getSession is used instead. 
**Mark** that getSession is async and to be used when user data is required

weihanwang: I'm not sure I understand your comment.

src/components/AppPageContainer.tsx
weihanwang: I don't see this comment is addressed.

UncleBu1728: I thought you meant remembering to solve this TODO in a separate PR since I mentioned the TODOs after this PR, sorry for the misunderstanding and this is resolved in the latest commit

src/components/AppPageContainer.tsx
UncleBu1728: The session was already properly passed by 
```javascript 
<SessionProvider session={session} />
``` 
So I meant that I do not need to address the `pass in the session object from _app.tsxLSwitchBoard()` mentioned in the previous comment, and you should just call `getSession()` API whenever the client side needs to use the user session data, I will make a note in this PR comment

src/api/auth.ts
```diff
@@ -26,19 +26,11 @@ export const authIntegration = () => middleware(async ({ ctx, next }) => {
...
-  invariant(email);
 
-  // TODO: extend next-auth session data to include other user fields, and remove this extra call to db
-  const user = await db.User.findOne({
-    where: { email },
-    attributes: userAttributes,
-  });
-  invariant(user);
+  const roles = session.user?.roles;
+  if (!roles || !isPermitted(roles, permitted)) throw forbiddenError();
```
weihanwang, line 31: `!roles || ` should not be needed. add `invariant(roles)` to before this line.

src/pages/api/auth/[...nextauth].ts
```diff
@@ -6,6 +6,15 @@ import { SendVerificationRequestParams } from "next-auth/providers";
...
+
+declare module "next-auth" {
```
Comment (weihanwang) [inline src/pages/api/auth/[...nextauth].ts:13]: can you add a comment explaining **why** we need this block? Note that it's not about **what** this block does.

src/components/AppPageContainer.tsx
weihanwang: 1. when useSession() is available, we should use it instead of getSession(). The former is the proper "React way" to do it.

2. The reason I am requesting passing in session from SwitchBoard() is to make the session object being not null as an invariant to AppPageContainer. Invariants are important for code maintainability.

weihanwang (approved):

weihanwang (changes requested):
please resolve the comment in AppPageContainer.tsx

src/components/AppPageContainer.tsx
```diff
@@ -6,23 +6,17 @@ import UserContext from "../UserContext";
...
 import NavBars, { sidebarBreakpoint, sidebarContentMarginTop, topbarHeight } from 'components/Navbars';
 import PageLoader from 'components/PageLoader';
-import { getSession } from 'next-auth/react';
+import { Session } from 'next-auth';
 
-export default function AppPageContainer({ children, wide, ...rest }: {
+export default function AppPageContainer({ children, wide, session, ...rest }: {
   wide: boolean
-} & PropsWithChildren) {
+} & PropsWithChildren & {session: Session}) {
```
weihanwang, line 13: Don't add another extension. list it after `wide`:

```
{
  wide: boolean,
  session: Session,
} & PropsWithChildren
```

src/components/AppPageContainer.tsx
```diff
@@ -6,23 +6,17 @@ import UserContext from "../UserContext";
...
-      if (session) {
-        setUser(session.user);
-      }
-    };
-
-    fetchSession();
-  }, []);
+    const f = async () => { setUser(session.user); };
+    f();
+  }, [session]);
```
weihanwang, line 19: remove this whole block

src/components/AppPageContainer.tsx
```diff
@@ -6,23 +6,17 @@ import UserContext from "../UserContext";
...
 import PageLoader from 'components/PageLoader';
-import { getSession } from 'next-auth/react';
+import { Session } from 'next-auth';
 
-export default function AppPageContainer({ children, wide, ...rest }: {
+export default function AppPageContainer({ children, wide, session, ...rest }: {
   wide: boolean
-} & PropsWithChildren) {
+} & PropsWithChildren & {session: Session}) {
   const [user, setUser] = useState<User | null>(null);
```
weihanwang, line 14: change this line to `const [user, setUser] = useState<User>(session.user);`

src/components/AppPageContainer.tsx
```diff
@@ -6,23 +6,17 @@ import UserContext from "../UserContext";
...
-      }
-    };
-
-    fetchSession();
-  }, []);
+    const f = async () => { setUser(session.user); };
+    f();
+  }, [session]);
 
   return !user ?
```
weihanwang, line 21: remove this `!user` conditional

src/pages/_app.tsx
```diff
@@ -50,9 +50,9 @@ function App({ Component, pageProps: { session, ...pageProps } }: {
 
 export default trpcNext.withTRPC(App);
 
-function SwitchBoard({ children, wide, ...rest }: PropsWithChildren & { wide: boolean }) {
+function SwitchBoard({ children, wide, ...rest }: PropsWithChildren & { wide: boolean}) {
   // TODO: combine what userSession().data returns and our own User object.
```
weihanwang, line 54: remove this TODO

weihanwang (approved):

PR #185, opened 2023-09-21
Fix Tencent Meeting Corner case

UncleBu1728:
A new condition is added to `updateOngoingMeeting()` to improve the corner case when a user creates a meeting but does not join right away. This could lead to the problem of users of the same group joining different meetings when other users of the same group create a meeting and overwrite the previously created link.

Now `updateOngoingMeeting()` will check if the meeting has been created for more than 3 minutes, and only destroy the meeting if it exists for more than 3 minutes and has no users in the meeting,

### DIFF
--- a/src/api/routes/meetings.ts
+++ b/src/api/routes/meetings.ts
@@ -7,12 +7,13 @@ import { createMeeting, getMeeting } from "../TencentMeeting";
 import { formatGroupName } from "shared/strings";
 import apiEnv from "api/apiEnv";
 import sleep from "../../shared/sleep";
-import {  notFoundError } from "api/errors";
+import { notFoundError } from "api/errors";
 import { emailRoleIgnoreError } from 'api/sendgrid';
 import sequelizeInstance from 'api/database/sequelizeInstance';
 import { checkPermissionForGroup } from './groups';
 import { groupAttributes, groupInclude } from 'api/database/models/attributesAndIncludes';
 import { Group } from '../../shared/Group';
+import moment from 'moment';
 
 /**
  * Find the group meeting by the input group id. If the meeting is presented in OngoingMeetings model, return the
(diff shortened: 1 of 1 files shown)


src/api/routes/meetings.ts
```diff
@@ -118,6 +119,11 @@ export async function updateOngoingMeetings() {
       attributes: ["tmUserId", "meetingId"], lock: true, transaction: t,
     });
     for (const meeting of ongoingMeetings) {
+      // This condition is added to improve the corner case when a user created a meeting but did not join
+      // If the user did not join meeting right away, other users of the same group might create a new meeting
+      // and overwrite the link. A 3-min grace period is added after a meeting is created to prevent such case.
+      if (moment().diff(meeting.createdAt, 'minutes') < 3) {continue;};
```
weihanwang, line 125: `{continue;};` => `continue;`

and I'm surprised the ts compiler didn't complain because there is no space between words and brackets.

src/api/routes/meetings.ts
```diff
@@ -23,61 +24,60 @@ import { Group } from '../../shared/Group';
...
-
-  // lock option is applied to transactions in order to prevent conflicts 
-  // when multiple transactions are trying to access or modify the same data simultaneously
-  // details: https://stackoverflow.com/a/48297781
-  // if the user's group is present in OngoingMeetings Model, return the meeting link 
-  return await sequelizeInstance.transaction(async (t) => {
-    const ongoingMeeting = await db.OngoingMeetings.findOne({ 
-      where: { groupId: g.id },
-      lock: true, transaction: t,
+  .mutation(async ({ ctx, input }) => {
```
weihanwang, line 27: please undo all unnecessary changes.

weihanwang (approved):

PR #197, opened 2023-09-27
Add handlebars to gen-test-data

UncleBu1728:
added `{{name}}` to `yarn gen-test-data` when generating summaries for testing name mapping API later

### DIFF
--- a/tools/generateTestData.ts
+++ b/tools/generateTestData.ts
@@ -45,7 +45,7 @@ async function main() {
   await upgradeUsers(mgrs);
   await generateUsers();
   await generateGroupsAndSummaries(mgrs);
-
+  // This make sure the process doesn't hang waiting for connection closure.
   await sequelize.close();
 }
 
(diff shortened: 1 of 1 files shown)


tools/generateTestData.ts
```diff
@@ -39,7 +39,7 @@ main().then();
 async function main() {
   // Force sequelize initialization
   const _ = sequelize;
```
weihanwang, line 41: we probably don't need these two lines

tools/generateTestData.ts
```diff
@@ -79,14 +79,16 @@ async function generateGroupsAndSummaries(include: User[]) {
 
   await generateSummaries([...include, mentees[1]]);
   await generateSummaries([...include, mentors[0]]);
+  // This make sure the process doesn't hang waiting for connection closure.
+  await sequelize.close();
```
weihanwang, line 83: good call, but don't put it here. put it at the end of main().

weihanwang (approved):

PR #198, opened 2023-09-27
Create name mapping model

UncleBu1728:
The name mapping model is associated with transcript IDs so it is unique per transcript and could apply to all the summaries under the associated transcript. Group IDs are not used since the group members could change.

### DIFF
--- a/src/api/database/db.ts
+++ b/src/api/database/db.ts
@@ -15,6 +15,7 @@ import Calibration from "./models/Calibration";
 import InterviewFeedbackUpdateAttempt from "./models/InterviewFeedbackUpdateAttempt";
 import ChatThread from "./models/ChatThread";
 import ChatMessage from "./models/ChatMessage";
+import NameMapping from "./models/TranscriptNameMapping";
 
 const db = {
   Partnership,
@@ -31,6 +32,7 @@ const db = {
   Calibration,
   ChatThread,
   ChatMessage,
+  NameMapping
 };
 
 export default db;
--- a/src/api/database/db.ts
+++ b/src/api/database/db.ts
@@ -15,15 +15,15 @@ import Calibration from "./models/Calibration";
 import InterviewFeedbackUpdateAttempt from "./models/InterviewFeedbackUpdateAttempt";
 import ChatThread from "./models/ChatThread";
 import ChatMessage from "./models/ChatMessage";
-import NameMapping from "./models/TranscriptNameMapping";
+import TranscriptNameMapping from "./models/TranscriptNameMapping";
 
 const db = {
   Partnership,
   User,
-  Group, 
-  GroupUser, 
-  Transcript, 
-  Summary, 
+  Group,
+  GroupUser,
+  Transcript,
+  Summary,
   OngoingMeetings,
   Assessment,
   Interview,
@@ -32,7 +32,7 @@ const db = {
   Calibration,
   ChatThread,
   ChatMessage,
-  NameMapping
+  TranscriptNameMapping,
 };
 
 export default db;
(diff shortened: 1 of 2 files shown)


src/api/database/models/TranscriptNameMapping.ts
```diff
@@ -0,0 +1,34 @@
+import {
```
weihanwang, line 1: indentation of the whole file is messed up.

src/api/database/models/TranscriptNameMapping.ts
```diff
@@ -0,0 +1,34 @@
...
+    Model,
+    ForeignKey,
+    BelongsTo,
+    PrimaryKey,
+  } from 'sequelize-typescript';
+  import { UUID, STRING } from 'sequelize';
+  import User from './User';
+  import Transcript from "./Transcript";
+
+  @Table({ paranoid: true })
```
weihanwang, line 13: we don't need paranoid mode it. (we should remove paranoid mode from most existing tables, too, in future prs).

src/api/database/models/TranscriptNameMapping.ts
```diff
@@ -0,0 +1,34 @@
...
+  class NameMapping extends Model {
+    @PrimaryKey
+    @Column(STRING)
+    handlebarName: string;
+
+    @PrimaryKey
+    @ForeignKey(() => Transcript)
+    @Column(STRING)
+    transcriptId: string;
```
weihanwang, line 22: no need for this field. name mapping should be global across all transcripts.

src/api/database/models/TranscriptNameMapping.ts
```diff
@@ -0,0 +1,34 @@
...
+    ForeignKey,
+    BelongsTo,
+    PrimaryKey,
+  } from 'sequelize-typescript';
+  import { UUID, STRING } from 'sequelize';
+  import User from './User';
+  import Transcript from "./Transcript";
+
+  @Table({ paranoid: true })
+  class NameMapping extends Model {
```
weihanwang, line 14: use full name `TranscriptNameMapping`

src/api/database/models/TranscriptNameMapping.ts
weihanwang, line 13: make a brief comment on what this class is about

src/api/database/models/TranscriptNameMapping.ts
```diff
@@ -0,0 +1,34 @@
...
+    handlebarName: string;
+
+    @PrimaryKey
+    @ForeignKey(() => Transcript)
+    @Column(STRING)
+    transcriptId: string;
+  
+    @Column(UUID)
+    @ForeignKey(() => User)
+    userId: string | null;
```
weihanwang, line 26: it should be non-null. use @AllowNull(false);

src/api/database/models/TranscriptNameMapping.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    // Associations
```
weihanwang, line 28: for consistency with other files, use

```
/**
 * Associations
 */

user

File: src/api/database/db.ts [hunk 7]
```diff
@@ -31,6 +32,7 @@ const db = {
   Calibration,
   ChatThread,
   ChatMessage,
+  NameMapping
```
weihanwang, line 35: add a comma

src/api/database/models/TranscriptNameMapping.ts
UncleBu1728: I was thinking if we should add any other field to identify the users in the case of different users having the same handlebar name

src/api/database/models/TranscriptNameMapping.ts
UncleBu1728: simply removed null should be sufficient, adding allowNull(false) would cause a column annotation error

weihanwang (approved):

weihanwang (approved):

PR #200, opened 2023-09-29
add handlebars api

UncleBu1728:
2 API are added to route transcript.ts
`getNameMap` will be called when the transcript component is used. It will get all the handlebars in the summaries under the same transcript, then create and return an object that maps the handlebars and names of the users. If there is no name map data for any of the handlebars, it will return the handlebar name.
`updateNameMap` is used for a modal that is similar to the mentorship page which updates and saves any name mapping selected by users


src/api/routes/transcripts.ts
```diff
@@ -52,8 +53,65 @@ const list = procedure
   return g.transcripts;
 });
 
+const getNameMap = procedure
```
weihanwang, line 56: Why not return translated summaries directly from summaries.list API?

src/api/routes/transcripts.ts
```diff
@@ -52,8 +53,65 @@ const list = procedure
...
+  .use(authUser())
+  .input(z.string())
+  .query(async ({ input: transcriptId }) => {
+
+    const summaries = await db.Summary.findAll({ where: { transcriptId } });
+
+    // find all handlebar from summaries under one transcript
+    let handlebars: string[] = [];
+    for (let s of summaries) {
+      const matches = s.summary.match(/{{(.*?)}}/g);
```
weihanwang, line 66: don't do it yourself. use an existing handlebar library.

src/api/routes/transcripts.ts
```diff
@@ -52,8 +53,65 @@ const list = procedure
...
+        invariant(nm.user.name);
+        nameMap[nm.handlebarName] = nm.user.name;
+      }
+    }
+
+    return nameMap;
+
+  });
+
+const updateNameMap = procedure
```
weihanwang, line 95: it doesn't belong here. create a new file (and router entry) `transcriptNameMapping.ts`.

also, use full names and avoid short names for symbols with large scope.

src/api/routes/summaries.ts
```diff
@@ -77,6 +78,42 @@ const list = procedure
...
+      nameMap[nm.handlebarName] = nm.user.name;
+    }
+  }
+  
+  // using Handlebars.js to compile and return summaries with handlebar replaced
+  t.summaries = t.summaries.map(summary => {
+    summary.summary = Handlebars.compile(summary.summary)(nameMap);
+    return summary;
+  });
+
```
weihanwang, line 116: Extract this whole block to smaller functions.

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,72 @@ const list = procedure
...
+    });
+
+    if (snm) {
+      for (const nm of snm) {
+        invariant(nm.user.name);
+        nameMap[nm.handlebarName] = nm.user.name;
+      }
+    }
+
+    return nameMap;
```
weihanwang, line 94: do not repeat yourself. extract common code of these routes into a common function.

src/api/routes/summaries.ts
```diff
@@ -77,6 +78,42 @@ const list = procedure
...
   checkPermissionForGroup(ctx.user, t.group);
 
+  // find all handlebar from summaries under one transcript
+  let handlebars: string[] = [];
+  for (let s of t.summaries) {
+    const matches = s.summary.match(/{{(.*?)}}/g);
+    // parse and find handlebar values but get rid of the repeating values
+    if (matches) {
+       handlebars = [...handlebars, ...matches.map(match => match.slice(2, -2).trim())]
+        .reduce((acc, cur) => acc.includes(cur) ? acc : [...acc, cur], handlebars);
```
weihanwang, line 88: this line is too cryptic. break it down into more readable smaller lines.

also, use this instead for simplicity:

````
const handlebar = ...

handlebars.push(...matches)
```

src/api/routes/summaries.ts
```diff
@@ -77,6 +78,42 @@ const list = procedure
...
+    }
+  };
+
+  let nameMap: Record<string, string> = {};
+  // create a list of namemap of itself, otherwise when handlebars compile it will return empty
+  nameMap = handlebars.reduce((o, key) => ({ ...o, [key]: key }), {});
```
weihanwang, line 94: this is too clever and not readable. use simpler / more readable way to achieve the same goal. for example, use a for loop

src/api/routes/summaries.ts
```diff
@@ -77,6 +78,42 @@ const list = procedure
...
+
+  const snm = await db.SummaryNameMapping.findAll({
+    where: { handlebarName: handlebars },
+    include: [{
+      model: db.User,
+      attributes: ['name']
```
weihanwang, line 100: need to add `where: { name is not null }` because user names can be null

src/api/routes/summaries.ts
```diff
@@ -77,6 +78,42 @@ const list = procedure
...
+    }],
+  });
+
+  if (snm) {
```
weihanwang, line 104: you don't need this line

src/api/routes/summaries.ts
```diff
@@ -77,6 +78,42 @@ const list = procedure
...
+    for (const nm of snm) {
+      invariant(nm.user.name);
+      nameMap[nm.handlebarName] = nm.user.name;
+    }
+  }
+  
+  // using Handlebars.js to compile and return summaries with handlebar replaced
+  t.summaries = t.summaries.map(summary => {
```
weihanwang, line 112: why not simply use a for loop

```
for (const s of t.suammries) s.suammry = Handlebars.compile(...);

File: src/api/routes/transcripts.ts [hunk 11]
```diff
@@ -52,8 +54,72 @@ const list = procedure
...
+
+  });
+
+const updateNameMap = procedure
+  .use(authUser())
+  .input(z.object({}).catchall(z.string()))
```
weihanwang, line 100: don't do this. use explicit type definition like `z.record(z.string(), ...)` instead

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,72 @@ const list = procedure
...
+  .mutation(async ({ input: nameMap }) => {
+    return sequelize.transaction(async (transaction) => {
```
weihanwang, line 102: replace `return` with `await`

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,72 @@ const list = procedure
...
+      // Construct an array of objects to upsert rows which userIds are not empty
+      const upsertArray = Object.entries(nameMap)
+        .filter(([handlebarName, userId]) => userId !== '')
```
weihanwang, line 105: front end should never pass in an invalid userId. it should omit the entry completely and not passing it down to the backend.

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,72 @@ const list = procedure
...
+        .map(([handlebarName, userId]) => ({
+          handlebarName,
+          userId,
+        }));
+
+      if (upsertArray.length > 0) {
```
weihanwang, line 111: this conditional is not necessary.

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,72 @@ const list = procedure
...
+        await db.SummaryNameMapping.bulkCreate(upsertArray, {
+          updateOnDuplicate: ['userId'], // specify the field(s) that should be updated on duplicate
```
weihanwang, line 113: this line is self explanatory so the comment on this line is not necessary.

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,72 @@ const list = procedure
...
+          transaction: transaction, // Use the transaction in bulkCreate
```
weihanwang, line 114: same here. remove comment. also, `transaction,` is sufficient

src/api/routes/summaries.ts
```diff
@@ -77,6 +79,14 @@ const list = procedure
 
   checkPermissionForGroup(ctx.user, t.group);
 
+  const handlebars = await extractHandlebars(transcriptId);
+  const nameMap = await createNameMap(handlebars);
+  
+  // using Handlebars.js to compile and return summaries with handlebar replaced
```
weihanwang, line 85: you don't need to use comments to describe WHAT the code does unless the code isn't too self explanatory.

src/api/routes/summaries.ts
```diff
@@ -77,6 +79,14 @@ const list = procedure
...
+  for (let summary of t.summaries) {
```
weihanwang, line 86: `const`

src/api/routes/summaries.ts
```diff
@@ -180,3 +190,48 @@ export async function findMissingCrudeSummaries(): Promise<CrudeSummaryDescripto
   }
   return ret;
 }
+
+export async function extractHandlebars(transcriptId: string): Promise<string[]> {
+  const summaries = await db.Summary.findAll({ where: { transcriptId } });
```
weihanwang, line 195: use `attributes` to retrieve only necessary fields

src/api/routes/summaries.ts
```diff
@@ -180,3 +190,48 @@ export async function findMissingCrudeSummaries(): Promise<CrudeSummaryDescripto
...
+
+  // find all handlebar from summaries under one transcript
+  let handlebars: string[] = [];
+  for (let s of summaries) {
+    const matches = s.summary.match(/{{(.*?)}}/g);
+    if (matches) {
+      // Extract matched handlebars and append to the existing handlebars
+      const extractedHandlebars = matches.map(match => match.slice(2, -2).trim());
```
weihanwang, line 203: why do you need trim here?

src/api/routes/summaries.ts
```diff
@@ -180,3 +190,48 @@ export async function findMissingCrudeSummaries(): Promise<CrudeSummaryDescripto
   }
   return ret;
 }
+
+export async function extractHandlebars(transcriptId: string): Promise<string[]> {
```
weihanwang, line 194: `extractHandlebarNames`

src/api/routes/summaries.ts
```diff
@@ -180,3 +190,48 @@ export async function findMissingCrudeSummaries(): Promise<CrudeSummaryDescripto
...
+        if (!handlebars.includes(extracted)) {
+          handlebars.push(extracted);
+        }
+      }
+    }
+  }
+  return handlebars;
+}
+
+export async function createNameMap(handlebars: string[]): Promise<Record<string, string>> {
```
weihanwang, line 215: `getNameMap`

src/api/routes/summaries.ts
```diff
@@ -180,3 +190,48 @@ export async function findMissingCrudeSummaries(): Promise<CrudeSummaryDescripto
...
+  let nameMap: Record<string, string> = {};
+  // create a list of namemap of itself, otherwise when Handlebars.js compile it will return empty
+  for (const handlebar of handlebars) {
+    nameMap[handlebar] = handlebar;
```
weihanwang, line 219:
```
... = `**${handlebar}**`
```

src/api/routes/summaries.ts
```diff
@@ -180,3 +190,48 @@ export async function findMissingCrudeSummaries(): Promise<CrudeSummaryDescripto
...
+    include: [{
+      model: db.User,
+      attributes: ['name'],
+      where: { name: { [Op.ne]: null } } // Ensure user's name is not null
+    }],
+  });
+
+  for (const nm of snm) {
+    invariant(nm.user.name);
+    nameMap[nm.handlebarName] = nm.user.name;
```
weihanwang, line 233:
```
... = `**${nm.user.name}**`
```

src/api/routes/summaries.ts
```diff
@@ -180,3 +190,48 @@ export async function findMissingCrudeSummaries(): Promise<CrudeSummaryDescripto
...
+    const matches = s.summary.match(/{{(.*?)}}/g);
+    if (matches) {
+      // Extract matched handlebars and append to the existing handlebars
+      const extractedHandlebars = matches.map(match => match.slice(2, -2).trim());
+      // Remove duplicates
+      for (const extracted of extractedHandlebars) {
+        if (!handlebars.includes(extracted)) {
+          handlebars.push(extracted);
+        }
+      }
```
weihanwang, line 209: use `Set` as the data type for `handlebars` so you don't need to dedupe manually

src/api/routes/summaries.ts
weihanwang, line 203: `extractedHandlebars` => `extracted`

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,37 @@ const list = procedure
   return g.transcripts;
 });
 
+const getNameMap = procedure
+  .use(authUser())
+  .input(z.string())
```
weihanwang, line 59: use `.input(z.object({ transcriptId: z.string() }})`. I think for newer code we should always specify the name of the parameter, only if there is only one parameter.

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,37 @@ const list = procedure
...
+  .query(async ({ input: transcriptId }) => {
```
weihanwang, line 60: specify `.output()`

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,37 @@ const list = procedure
...
+    const handlebars = await extractHandlebars(transcriptId);
+    return await createNameMap(handlebars);
+  });
+
+const updateNameMap = procedure
```
weihanwang, line 65: comment on the expected input

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,37 @@ const list = procedure
...
+  .use(authUser())
+  .input(z.record(z.string()))
+  .mutation(async ({ input: nameMap }) => {
+    await sequelize.transaction(async (transaction) => {
+      // Construct an array of objects to upsert rows which userIds are not empty
```
weihanwang, line 70: this comment is not accurate.

src/api/routes/transcripts.ts
```diff
@@ -52,8 +54,37 @@ const list = procedure
...
+  .query(async ({ input: transcriptId }) => {
+    const handlebars = await extractHandlebars(transcriptId);
+    return await createNameMap(handlebars);
+  });
+
+const updateNameMap = procedure
+  .use(authUser())
+  .input(z.record(z.string()))
+  .mutation(async ({ input: nameMap }) => {
+    await sequelize.transaction(async (transaction) => {
```
weihanwang, line 69: I don't think you need a transaction here because there is only one sql statement.

src/api/routes/summaries.ts
weihanwang, line 194: move these functions into transcripts.ts because they are more related to transcripts than summaries.

weihanwang, line 194: if you always call extrcthandlebars and createNameMap together, either create a wrapper function so callers only call this single wrapper function (and do not export the other two functions), or merge these two functions into one.

src/api/routes/summaries.ts
UncleBu1728: I have trim here to get rid of the curly brackets `{{ }}` wrapped around the handlebars

src/api/routes/transcripts.ts
```diff
@@ -52,8 +59,74 @@ const list = procedure
...
+    attributes: summaryAttributes,
+  });
+
+  // find all handlebar from summaries under one transcript
+  let handlebarsSet = new Set<string>();
+
+  for (let s of summaries) {
+    const matches = s.summary.match(/{{(.*?)}}/g);
+    if (matches) {
+      // trim is to remove curly brackets wrapped around handlebars found by regex
```
weihanwang, line 103: `slice(2, -2)` already removes the brackets. I still don't see why trim is necessary.

src/api/routes/transcripts.ts
```diff
@@ -52,8 +59,74 @@ const list = procedure
...
+      const extracted = matches.map(match => match.slice(2, -2).trim());
+      extracted.forEach(handlebar => handlebarsSet.add(handlebar));
+    }
+  }
+
+  let handlebars = [...handlebarsSet];
```
weihanwang, line 109: inline this variable

src/api/routes/transcripts.ts
```diff
@@ -52,8 +59,74 @@ const list = procedure
...
   get,
   list,
+  getNameMap,
+  updateNameMap
 });
+
+export async function extractAndGetNameMap(transcriptId: string): Promise<Record<string, string>> {
+  const summaries = await db.Summary.findAll({
+    where: { transcriptId },
+    attributes: summaryAttributes,
```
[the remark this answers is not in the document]
UncleBu1728: if this function will be used to return summary.list, `summaryKey` is also requried for frontend.

src/api/routes/transcripts.ts
```diff
@@ -52,8 +59,74 @@ const list = procedure
...
+});
 
 export default router({
   get,
   list,
+  getNameMap,
+  updateNameMap
 });
+
+export async function extractAndGetNameMap(transcriptId: string): Promise<Record<string, string>> {
```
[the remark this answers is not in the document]
UncleBu1728: Please see the comment below. For the latest commit, I have assign the summary list to the transcript object as you suggested, but I think we could just use the returned `summaries` from this function directly instead of assigning it to the object then return.

src/api/routes/transcripts.ts
```diff
@@ -52,8 +59,76 @@ const list = procedure
...
+    attributes: summaryAttributes,
+  });
+
+  // find all handlebar from summaries under one transcript
+  let handlebarsSet = new Set<string>();
+
+  for (let s of summaries) {
+    const matches = s.summary.match(/{{(.*?)}}/g);
+    if (matches) {
+      // trim is to remove curly brackets wrapped around handlebars found by regex
```
weihanwang, line 107: need to remove this comment

src/api/routes/transcripts.ts
```diff
@@ -52,8 +59,76 @@ const list = procedure
...
 });
+
+export async function getSummariesAndNameMap(transcriptId: string): Promise<{ summaries: typeof summaries, nameMap: typeof nameMap }> {
+  const summaries = await db.Summary.findAll({
+    where: { transcriptId },
+    attributes: summaryAttributes,
+  });
+
+  // find all handlebar from summaries under one transcript
+  let handlebarsSet = new Set<string>();
```
weihanwang, line 102: const

weihanwang (approved):

PR #210, opened 2023-10-10
Add NameMapping Frontend

UncleBu1728:
[Demo1](https://drive.google.com/file/d/1dmroK5flYOUS5hafFnusthyOJsyIuVJ8/view)
[Demo2](https://drive.google.com/file/d/1pFkZLw64nL-DI14Tu6gxqgyxVxB108qg/view)
Below is a screenshot of options in select due to was not captured by screen record
ea7ab51996053dd71bfc928e8b350b0


src/shared/Summary.ts
```diff
@@ -7,3 +7,7 @@ export const zSummary = z.object({
   });
 
 export type Summary = z.TypeOf<typeof zSummary>;
+
+export const zNameMap = z.record(z.string());
```
weihanwang, line 11: Where do you use these variables? Also, again, use full name: zSummaryNameMap

src/components/Transcripts.tsx
```diff
@@ -85,19 +106,128 @@ function LoadedTranscripts({ transcripts: unsorted }: {
...
+          onClose={() => {
+            handleNameMapModal();
+            refetchSummary();
+            refetchNameMap();
+          }} />}
     </Flex>
   );
 }
+
+function NameMapModal({ nameMap, groupUsers, onClose }: {
```
weihanwang, line 136: `SummaryNameMapModal`

Also, change models.SummaryNameMapping class to models.SummaryNameMap, but in a separate PR.

weihanwang, line 136: Move this function into a separate file under the components folder. Also break this function down to smaller functions. It's too long as it stands right now.

src/components/Transcripts.tsx
weihanwang, line 1: There are build errors. Please fix.

src/shared/Summary.ts
UncleBu1728: The SummaryNameMap type is used to define the nameMap variable in nameMapModal

src/components/SummaryNameMapModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+} from '@chakra-ui/react';
+import React, { useEffect, useState } from 'react';
+import trpc from "../trpc";
+import ModalWithBackdrop from 'components/ModalWithBackdrop';
+import ReactMarkdown from 'react-markdown';
+import { MinUser } from 'shared/User';
+import { SummaryNameMap } from 'shared/Summary';
+
+export function SummaryNameMapModal({ nameMap, groupUsers, onClose }: {
+    nameMap: SummaryNameMap | undefined
```
weihanwang, line 22: when nameMap is expected to be undefined? document it:

```
/**
 * @param nameMap ...
 */
```

src/components/SummaryNameMapModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+    groupUsers: MinUser[]
+    onClose: () => void,
+}) {
+    // UserMap = { [handlebar] : { name: [userName], id: [userId] }}
+    // May update summaryNameMap API to output UserMap instead
```
weihanwang, line 27: I don't quite understand what this sentence

src/components/SummaryNameMapModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+    const [updatedUserMap, setUpdatedUserMap] = useState<{ [key: string]: SummaryNameMap }>({});
+
+    useEffect(() => {
```
weihanwang, line 30: No need for useEffect. the computation isn't that expensive.

src/components/SummaryNameMapModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+import ReactMarkdown from 'react-markdown';
+import { MinUser } from 'shared/User';
+import { SummaryNameMap } from 'shared/Summary';
+
+export function SummaryNameMapModal({ nameMap, groupUsers, onClose }: {
+    nameMap: SummaryNameMap | undefined
+    groupUsers: MinUser[]
+    onClose: () => void,
+}) {
+    // UserMap = { [handlebar] : { name: [userName], id: [userId] }}
```
weihanwang, line 26: okay I think you will need to change the `getNameMap` to return `SummaryNameMap` object which contains userids and not user names. This will greatly simplify your code and makes more sense.

weihanwang: also change the parameter type of `updateNameMap` to `SummaryNameMap`. All these backend changes should be done in a separate PR.

src/components/SummaryNameMapModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+        }
+    };
+
+    return (
+        <ModalWithBackdrop isOpen onClose={onClose}>
+            <ModalOverlay />
+            <ModalContent>
+                <ModalHeader>用户匹配</ModalHeader>
+                <ModalCloseButton />
+                <FormControl>
```
weihanwang, line 82: Why do you need Form control here?

src/components/SummaryNameMapModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+            onClose();
+        }
+    };
+
+    return (
+        <ModalWithBackdrop isOpen onClose={onClose}>
+            <ModalOverlay />
+            <ModalContent>
+                <ModalHeader>用户匹配</ModalHeader>
+                <ModalCloseButton />
```
weihanwang, line 81: shouldn't it within ModelHeader?

src/components/Transcripts.tsx
```diff
@@ -98,6 +111,15 @@ function LoadedTranscripts({ transcripts: unsorted }: {
...
         </Card>
       }
+      {nameMapModal &&
+        <SummaryNameMapModal
+          nameMap={nameMap}
+          groupUsers={groupUsers}
+          onClose={() => {
+            handleNameMapModal();
+            refetchSummary();
+            refetchNameMap();
```
weihanwang, line 121: no need for fetch calls. use trpcNext's invalidate method from within the `save` function in `SummaryNameMapModal`. Search the codebase for "invalidate" for references.

src/api/routes/summaries.ts
```diff
@@ -80,9 +80,16 @@ const list = procedure
   checkPermissionForGroup(ctx.user, t.group);
 
   const { nameMap, summaries } = await getSummariesAndNameMap(transcriptId);
-
-  for (const summary of summaries) {
-    summary.summary = Handlebars.compile(summary.summary)(nameMap);
+  
+  try {
```
weihanwang, line 84: put try inside for loop so not all summaries are affected by malformed ones.

src/components/SummaryNameMapModal.tsx
UncleBu1728: a warning is added instead when nameMap is undefined

src/components/SummaryNameMapModal.tsx
UncleBu1728: useEffect is needed for the page to render, otherwise will return blank page

src/components/SummaryNameMapModal.tsx
UncleBu1728: `FormControl` is needed for organizing form elements  `FormLabel, Input` and `Select`, otherwise it will yell 
the error `JSX expressions must have one parent element.`
You could resolve this using an empty expression but the format will be messed up

UncleBu1728:
[Demo 3](https://drive.google.com/file/d/1c1PHsGv6FNworIRqvOR_VBDNoGyBz-5g/view) - current

image

Demo 3 should be the same as Demo2, just with the latest modification in codes

PR #215, opened 2023-10-15
modify handlebar backend api

UncleBu1728:
- The APIs and functions named using `NameMap` are changed to `UserMap` for a more accurate description. 
- UserMap consists of a `zSummaryNameMap` type of `{ [handlebar]:  string, [userId]: string, [user]: MinUser }` which will be passed and used at the frontend. 
- `NameMap` is retained for its utility with summary.list. This ensures compatibility with handlebar.js when rendering summary texts.

### DIFF
--- a/src/api/routes/transcripts.ts
+++ b/src/api/routes/transcripts.ts
@@ -55,11 +55,11 @@ const getMostRecentStartedAt = procedure
 const getNameMap = procedure
   .use(authUser())
   .input(z.object({ transcriptId: z.string() }))
-  .output(z.array(zTranscriptNameMap))
+  .output(zTranscriptNameMap)
   .query(async ({ input }) =>
 {
   const { nameMap } = await getSummariesAndNameMap(input.transcriptId);
-  return [nameMap];
+  return nameMap;
 });
 
 /**
@@ -114,22 +114,27 @@ export async function getSummariesAndNameMap(transcriptId: string): Promise<{
     where: { handlebarName: [...handlebarsSet] },
     include: [{
       model: db.User,
-      attributes: ['name'],
+      attributes: ['name','id'],
       where: { name: { [Op.ne]: null } } // Ensure user's name is not null
     }],
   });
 
-  let nameMap: TranscriptNameMap = tnm;
-  
+  let nameMap: TranscriptNameMap = [];
   // check if the handlebar names are presented in the returned array of transcriptNameMap
   // if not, create an object of this handlebar with null user, 
   // otherwise when Handlebars.js compile it will return empty
   for (const handlebar of [...handlebarsSet]) {
-    if (![...nameMap].some(nm => nm.handlebarName === handlebar)) {
+    const entries = tnm.filter(e => e.handlebarName === handlebar);
+    if (entries.length) {
+      const mappedEntries = entries.map(e => ({
+        handlebarName: e.handlebarName,
+        user: e.user.dataValues
+      }));
+      nameMap.push(...mappedEntries);
+    } else {
       nameMap.push({ handlebarName: handlebar, user: null });
     }
   }
 
-  console.log(nameMap);
   return { summaries, nameMap };
 }
--- a/src/api/routes/transcripts.ts
+++ b/src/api/routes/transcripts.ts
@@ -11,7 +11,6 @@ import {
   summaryAttributes
 } from "api/database/models/attributesAndIncludes";
 import { checkPermissionForGroup } from "./groups";
-import invariant from 'tiny-invariant';
 import { Op } from "sequelize";
 import Summary from "api/database/models/Summary";
 import { zTranscriptNameMap, TranscriptNameMap } from "shared/Transcript";
@@ -53,29 +52,29 @@ const getMostRecentStartedAt = procedure
   return await db.Transcript.max("startedAt", { where: { groupId } });
 });
 
-const getUserMap = procedure
+const getNameMap = procedure
   .use(authUser())
   .input(z.object({ transcriptId: z.string() }))
   .output(z.array(zTranscriptNameMap))
   .query(async ({ input }) =>
 {
-  const { userMap } = await getSummariesAndUserMap(input.transcriptId);
-  return [userMap];
+  const { nameMap } = await getSummariesAndNameMap(input.transcriptId);
+  return [nameMap];
 });
 
 /**
  * @param { [handlebarNames]: userIds }
  */
-const updateUserMap = procedure
+const updateNameMap = procedure
   .use(authUser())
   .input(z.array(z.object({
     handlebarName: z.string(),
     userId: z.string(),
   })))
-  .mutation(async ({ input: userMap }) =>
+  .mutation(async ({ input: nameMap }) =>
 {
   // Construct an array of objects to upsert multiple rows the same time
-  const upsertArray = Object.entries(userMap)
+  const upsertArray = Object.entries(nameMap)
     .map(([handlebarName, userId]) => ({
       handlebarName,
       userId,
@@ -87,14 +86,13 @@ const updateUserMap = procedure
 export default router({
   list,
   getMostRecentStartedAt,
-  getUserMap,
-  updateUserMap,
+  getNameMap,
+  updateNameMap,
 });
 
-export async function getSummariesAndUserMap(transcriptId: string): Promise<{
+export async function getSummariesAndNameMap(transcriptId: string): Promise<{
   summaries: Summary[],
   nameMap: TranscriptNameMap,
-  userMap: typeof snm,
 }> {
   const summaries = await db.Summary.findAll({
     where: { transcriptId },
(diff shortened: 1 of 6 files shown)


src/shared/Summary.ts
```diff
@@ -7,3 +8,9 @@ export const zSummary = z.object({
   });
 
 export type Summary = z.TypeOf<typeof zSummary>;
+
+export const zSummaryNameMap = z.object({
+  handlebarName: z.string(),
+  userId: z.string(),
```
weihanwang, line 14: why you need userId if there is already a user field?

src/api/routes/transcripts.ts
```diff
@@ -52,26 +53,26 @@ const getMostRecentStartedAt = procedure
   return await db.Transcript.max("startedAt", { where: { groupId } });
 });
 
-const getNameMap = procedure
+const getUserMap = procedure
```
weihanwang, line 56: Why renamed to user name? it should be consistent with zSummaryNameMap. So please use `getNameMap`.

src/api/routes/transcripts.ts
```diff
@@ -52,26 +53,26 @@ const getMostRecentStartedAt = procedure
...
-  return nameMap;
+  const { userMap } = await getSummariesAndUserMap(input.transcriptId);
+  return userMap;
 });
 
 /**
  * @param { [handlebarNames]: userIds }
  */
-const updateNameMap = procedure
+const updateUserMap = procedure
```
weihanwang, line 69: `updateNameMap`

src/api/routes/transcripts.ts
```diff
@@ -52,26 +53,26 @@ const getMostRecentStartedAt = procedure
...
   .use(authUser())
-  .input(z.record(z.string()))
-  .mutation(async ({ input: nameMap }) => 
+  .input(z.array(zSummaryNameMap))
+  .mutation(async ({ input: userMap }) =>
```
weihanwang, line 72: `nameMap`

src/api/routes/transcripts.ts
```diff
@@ -52,26 +53,26 @@ const getMostRecentStartedAt = procedure
...
 
 /**
  * @param { [handlebarNames]: userIds }
  */
-const updateNameMap = procedure
+const updateUserMap = procedure
   .use(authUser())
-  .input(z.record(z.string()))
-  .mutation(async ({ input: nameMap }) => 
+  .input(z.array(zSummaryNameMap))
```
weihanwang, line 71: don't use zSummaryNameMap here. simply

```
z.array(z.object({
  handlebarName: z.string(),
  userId: z.string(),
}))
```

src/shared/Summary.ts
```diff
@@ -7,3 +8,9 @@ export const zSummary = z.object({
   });
 
 export type Summary = z.TypeOf<typeof zSummary>;
+
+export const zSummaryNameMap = z.object({
```
weihanwang, line 12: move to Transcript.ts

weihanwang, line 12: define it as an array:`zTranscriptNameMap = z.array(z.object(...`

src/api/routes/transcripts.ts
weihanwang, line 56: Because the name map is more closely associated with transcripts than summaries, I suggest we change zSumaryNameMap to zTranscriptNameMap, and also rename the SummaryNameMap model.

src/api/routes/transcripts.ts
```diff
@@ -83,13 +84,20 @@ const updateNameMap = procedure
...
 });
 
-export async function getSummariesAndNameMap(transcriptId: string): Promise<{
+/**
+ * Retrieves summaries and generates name and ID mappings based on the handlebars within those summaries.
+ * @returns {Summary[]} result.summaries - An array of fetched summaries.
+ * @returns {Record<string, string>} result.nameMap - A mapping from handlebars to names.
+ * @returns {SummaryNameMap[]} result.userMap - A mapping from handlebars to user IDs and names.
+ */
+export async function getSummariesAndUserMap(transcriptId: string): Promise<{
```
weihanwang, line 97: `getSummariesAndNameMap`

src/api/routes/transcripts.ts
```diff
@@ -83,13 +84,20 @@ const updateNameMap = procedure
...
   summaries: Summary[],
   nameMap: Record<string, string>,
+  userMap: typeof snm,
```
weihanwang, line 100: don't return two fields. return only one: `nameMap: TranscriptNameMap`

src/shared/Summary.ts
```diff
@@ -7,3 +8,9 @@ export const zSummary = z.object({
...
+  handlebarName: z.string(),
+  userId: z.string(),
+  user: zMinUser,
+});
```
weihanwang, line 16: add `TranscriptNameMap = z.TypeOf<typeof zTranscriptNameMap>;`

src/api/routes/transcripts.ts
```diff
@@ -83,13 +84,20 @@ const updateNameMap = procedure
...
+  getUserMap,
+  updateUserMap,
 });
 
-export async function getSummariesAndNameMap(transcriptId: string): Promise<{
+/**
+ * Retrieves summaries and generates name and ID mappings based on the handlebars within those summaries.
+ * @returns {Summary[]} result.summaries - An array of fetched summaries.
+ * @returns {Record<string, string>} result.nameMap - A mapping from handlebars to names.
+ * @returns {SummaryNameMap[]} result.userMap - A mapping from handlebars to user IDs and names.
```
weihanwang, line 95: These return values are self evident, so no need these comments.

src/api/routes/transcripts.ts
UncleBu1728: So there are two objects returning from the `getSummariesAndUserMap` function, one is { [handlebar:] [user name] }, which is what we initially called `Name Map`. And there is also another object  `{ [handlebar:] [user] }` which is referred to as `User Map` here now returned to the frontend in order to make the frontend code work more simply. `Name Map` is still kept to be passed to handlebars.js to compile

src/api/routes/transcripts.ts
```diff
@@ -122,10 +119,17 @@ export async function getSummariesAndNameMap(transcriptId: string): Promise<{
     }],
   });
 
-  for (const nm of snm) {
-    invariant(nm.user.name);
-    nameMap[nm.handlebarName] = `**${nm.user.name}**`;
+  let nameMap: TranscriptNameMap = tnm;
```
weihanwang, line 122: This is the wrong way to do it. line 129 is changing `tnm` directly which is inappropriate. Use this instead:

```
const nameMap: TranscriptNameMap = [];

for (const handlebar of [...handlebarsSet]) {
  const entries = nameMap.filter(e => e.handlebarName === handlebar);
  nameMap.push(entries.length ? ... : ...);
}
```

src/api/routes/transcripts.ts
```diff
@@ -122,10 +119,17 @@ export async function getSummariesAndNameMap(transcriptId: string): Promise<{
...
+  // check if the handlebar names are presented in the returned array of transcriptNameMap
+  // if not, create an object of this handlebar with null user, 
+  // otherwise when Handlebars.js compile it will return empty
+  for (const handlebar of [...handlebarsSet]) {
+    if (![...nameMap].some(nm => nm.handlebarName === handlebar)) {
+      nameMap.push({ handlebarName: handlebar, user: null });
+    }
   }
 
+  console.log(nameMap);
```
weihanwang, line 133: remove

src/shared/Transcript.ts
```diff
@@ -7,3 +8,10 @@ export const zTranscript = z.object({
 });
 
 export type Transcript = z.TypeOf<typeof zTranscript>;
+
+export const zTranscriptNameMap = z.array(z.object({
+  handlebarName: z.string(),
+  user: zMinUser.nullable(),
```
weihanwang, line 14: comment on when it will be null

src/api/routes/transcripts.ts
```diff
@@ -114,22 +114,27 @@ export async function getSummariesAndNameMap(transcriptId: string): Promise<{
...
-  
+  let nameMap: TranscriptNameMap = [];
   // check if the handlebar names are presented in the returned array of transcriptNameMap
   // if not, create an object of this handlebar with null user, 
   // otherwise when Handlebars.js compile it will return empty
   for (const handlebar of [...handlebarsSet]) {
-    if (![...nameMap].some(nm => nm.handlebarName === handlebar)) {
+    const entries = tnm.filter(e => e.handlebarName === handlebar);
+    if (entries.length) {
+      const mappedEntries = entries.map(e => ({
```
weihanwang, line 129: don't use `map`. do

```
invariant(entries.length == 1);
manMap.push( ...entries[0]... );
```

src/api/routes/transcripts.ts
```diff
@@ -114,22 +114,27 @@ export async function getSummariesAndNameMap(transcriptId: string): Promise<{
     where: { handlebarName: [...handlebarsSet] },
     include: [{
       model: db.User,
-      attributes: ['name'],
+      attributes: ['name','id'],
```
weihanwang, line 117: use `minUserAttributes`

src/api/routes/transcripts.ts
UncleBu1728: image
image

the returned array of objects is shown above. So an object with specifically assigned values was pushed instead a spread operator

weihanwang (approved):

PR #216, opened 2023-10-16
Fix handlebar corner case 

UncleBu1728:
- A try-catch block is added to the `summaries.list` to return the original summary text if handlebars.js is having errors compiling.
- The regex method to find handlebars are replaced with a stack method

### DIFF
--- a/src/api/routes/summaries.ts
+++ b/src/api/routes/summaries.ts
@@ -81,8 +81,15 @@ const list = procedure
 
   const { nameMap, summaries } = await getSummariesAndNameMap(transcriptId);
 
-  for (const summary of summaries) {
-    summary.summary = Handlebars.compile(summary.summary)(nameMap);
+  try {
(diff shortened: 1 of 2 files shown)


src/api/routes/summaries.ts
```diff
@@ -81,15 +81,14 @@ const list = procedure
...
 
-  try {
-    for (const summary of summaries) {
+  for (const summary of summaries) {
+    try {
       // Compile and update summary
       summary.summary = Handlebars.compile(summary.summary)(nameMap);
+    } catch (error) {
+      // If there's an error compiling, keep and return the original summaries
+      console.error("Error compiling Handlebars template for summary:", summary.transcriptId + summary.summaryKey);
```
weihanwang, line 90: replace `+` with `,`

weihanwang (approved):

PR #220, opened 2023-10-31
handle 500 error and add sentry tracking

UncleBu1728:
a try-catch block is added to handle the cron job 500 error
So if the `meeting_id`s does not return valid outputs, Sentry should log the error in the catch block and the function would return an empty array for the `summaries/findMissingCrudeSummries` to work as normally


UncleBu1728:
In commit 5d9e9ce, a function is added to validate the HTTP response from Tencent Meeting API first to see if it is success before

UncleBu1728:
In the latest commit, `Axios` is used instead of the `requestWithBody` method due to the reason that adding `Axios` in a different PR would need to erase and rewrite the error handler added in the previous commit. This commit is using Axios to access Tencent Meeting RestAPI and will handle errors such as sending invalid requests that will cause Tencent RestAPI to return error responses or receiving an undefined object that will not pass Zod schema validation

PR #235, opened 2024-03-12
add sequelize-cli dependency and config

UncleBu1728:
This PR is to add [Sequelize Command-Line Interface (CLI)](https://github.com/sequelize/cli) to manage and keep track of changes to database data


models/index.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
```
weihanwang, line 1: why not modify src/api/database/sequelize.js directly? Also the JS syntax of this file is inconsistent with the rest of the source files.

config/sequelize-cli-config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
weihanwang, line 1: why do we need this block?

models/index.js
UncleBu1728: This is auto-generated by Sequelize Cli after run `npx sequelize-cli init`

config/sequelize-cli-config.js
UncleBu1728: This is the configuration of Sequelize cli which was also generated after running `npx sequelize-cli init`, a module is used instead of [a JSON shown in the document ](https://sequelize.org/docs/v6/other-topics/migrations/#configuration)in order to all the environment variables

weihanwang: 1. add a comment to explain the reason for this block.
2. didn't we say to replace the DATABASE_URL with user, password, etc?

weihanwang (approved):

PR #239, opened 2024-03-24
Set up login for e2e test and add example test

Linco1n3:


cypress/README.md
```diff
@@ -15,6 +15,29 @@ yarn cypress-run
 
 For more usage, refer to [Cypress Commands](https://docs.cypress.io/guides/guides/command-line).
 
+### Setup your Cypress env variables
+- the Cypress suite requires an environment variable that should be stored in your `.env.development` and not committed to git.
```
weihanwang, line 19: can it be .env.local instead of .env.development?

Linco1n3: actually was about to use .env.local. Updated.

weihanwang (approved):

PR #271, opened 2024-05-11
Add markdown styler to correctly render Markdown texts using Chakra UI

UncleBu1728:
The Chakra UI CSSReset component would override the default styles of elements provided by the browsers, which could cause markdown texts not to display correctly as shown below.

image

A markdown styler component has been added with customized component styles to resolve this problem.

2f919868fcbb71cde109ef5aa190c07

### DIFF
--- a/src/components/MarkdownStyler.tsx
+++ b/src/components/MarkdownStyler.tsx
@@ -1,0 +1,14 @@
+import React from 'react';
+import ReactMarkdown from 'react-markdown';
+import styles from '../theme/MarkdownStyler.module.css';
+
+// This component is to correctly render Markdowns when using chakra UI
+// Chakra UI CSSReset component is turning down browser default style of elements
+// https://stackoverflow.com/a/64317290
+const MarkdownStyler = ({ content }: { content: string }) => (
+  <div className={styles.markdownStyler}>
+    <ReactMarkdown>{content}</ReactMarkdown>
+  </div>
+);
+
+export default MarkdownStyler;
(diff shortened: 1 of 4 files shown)


src/components/MarkdownStyler.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import ReactMarkdown from 'react-markdown';
+import styles from '../theme/MarkdownStyler.module.css';
+
+// This component is to correctly render Markdowns when using chakra UI
+// Chakra UI CSSReset component is turning down browser default style of elements
```
weihanwang, line 6: Can you add a reference link about this?

src/theme/MarkdownStyler.module.css
```diff
@@ -0,0 +1,97 @@
...
+  
+  .markdownStyler h5 {
+    font-size: 1.1rem;
+  }
+  
+  .markdownStyler h6 {
+    font-size: 1rem;
+  }
+  
+  .markdownStyler p {
```
weihanwang, line 45: Please add a 1/2 line height of vertical spacing between <p>s

src/theme/MarkdownStyler.module.css
```diff
@@ -0,0 +1,97 @@
...
+    margin: 1rem 0;
+    line-height: 1.5;
+  }
+  
+  .markdownStyler del {
```
weihanwang, line 50: Can remove this line and all the lines below here. We don't need them for now.

weihanwang (changes requested):

Issue #272, opened 2024-05-11
React Markdown not recognizing strikethrough


UncleBu1728:
![2f919868fcbb71cde109ef5aa190c07](https://github.com/yuanjian-org/app/assets/83266861/09aa1fa3-8dbe-4426-9ff8-78380940366a)

React Markdown does not recognize "~~ word ~~" as strikethrough ~~word~~.  
The words are not wrapped in "< del >" or "< s >" shown on web page HTML

PR #273, opened 2024-05-11
Comment on remark-gfm to support GitHub Flavored Markdown(GFM)

UncleBu1728:
[issue 272](https://github.com/yuanjian-org/app/issues/272) reported that strikethrough is not recognized correctly. 

According to react-markdown official, the original dependency does not support strikethrough, so the plugin [remark-gfm](https://github.com/remarkjs/remark-gfm) is added and used as [shown in the README.](https://github.com/remarkjs/react-markdown?tab=readme-ov-file#use)

Now strikethrough is working properly as shown below

### DIFF
--- a/src/components/MarkdownStyler.tsx
+++ b/src/components/MarkdownStyler.tsx
@@ -5,6 +5,10 @@ import styles from '../theme/MarkdownStyler.module.css';
 // This component is to correctly render Markdowns when using chakra UI
 // Chakra UI CSSReset component is turning down browser default style of elements
 // https://stackoverflow.com/a/64317290
+//
+// To enable and render GFM such as strikethrough or footnote in the future
+// Add & use remark-gfm as plugin shown in the link below
+// https://github.com/remarkjs/react-markdown?tab=readme-ov-file#use
 const MarkdownStyler = ({ content }: { content: string }) => (
   <div className={styles.markdownStyler}>
     <ReactMarkdown>{content}</ReactMarkdown>


src/components/MarkdownStyler.tsx
```diff
@@ -1,13 +1,14 @@
...
 import styles from '../theme/MarkdownStyler.module.css';
+import remarkGfm from 'remark-gfm';
 
 // This component is to correctly render Markdowns when using chakra UI
 // Chakra UI CSSReset component is turning down browser default style of elements
 // https://stackoverflow.com/a/64317290
 const MarkdownStyler = ({ content }: { content: string }) => (
   <div className={styles.markdownStyler}>
-    <ReactMarkdown>{content}</ReactMarkdown>
+    <ReactMarkdown remarkPlugins={[remarkGfm]}>{content}</ReactMarkdown>
```
weihanwang, line 11: Add a comment on why we need remarkGfm.

UncleBu1728:
> [issue 272](https://github.com/yuanjian-org/app/issues/272) reported that strikethrough is not recognized correctly.
> 
> According to react-markdown official, the original dependency does not support strikethrough, so the plugin [remark-gfm](https://github.com/remarkjs/remark-gfm) is added and used as [shown in the README.](https://github.com/remarkjs/react-markdown?tab=readme-ov-file#use)
> 
> Now strikethrough is working properly as shown below 

"remark-gfm" is not added to the project at the moment since there is no requirement for its usage. However, a comment has been added in the codebase in case it is needed in the future.

Issue #92, opened 2023-07-15
Toast fails to show error message when attempting API calls to the server during network disconnection


weihanwang:
Browser console shows the following:

<img width="611" alt="image" src="https://github.com/yuanjian-org/app/assets/9847625/bc722549-e8fc-43ff-a846-af1ca94f7ccd">

Adagggggg:
The previously reported issue with the toast error message during network disconnections has been resolved already.

Currently, when the network disconnects, if we attempt to update a user, the appropriate error message is displayed as intended:
<img width="1468" alt="disconnect-error" src="https://github.com/yuanjian-org/app/assets/122472773/bcdaefc8-8568-4d82-ad7c-ffd19192cf7d">

Additionally, the spinner error has also already been fixed. The error related to `Cannot read properties of undefined` in the issue screenshot might be caused by [line 32 in trpc.ts](https://github.com/yuanjian-org/app/blob/main/src/trpc.ts#L32). If we remove the question mark of current code `err.data?.code`, a spinner will show up when updating the user during network disconnection:
<img width="1466" alt="code" src="https://github.com/yuanjian-org/app/assets/122472773/f7068dba-fe0f-43f4-bd28-d40c26eea177">

I suggest this issue can be closed. @weihanwang

weihanwang:
@Adagggggg 这个bug只有在登录界面上才会复现。

可以这样复现：

1. 在本地退出登录，回到 http://localhost:3000/auth/login 这一页。

2. 然后停掉后台的 postgres db。比如，如果你用的是 mac 和 homebrew，在terminal运行这个命令：$ brew services stop postgresql

3. 在登录界面输入email以后点击“登录 / 注册”

Adagggggg:
Fixed in [pr#279 ](https://github.com/yuanjian-org/app/pull/279)

PR #279, opened 2024-05-16
fix: fix toast error fails to show during network disconnection on login

Adagggggg:
This issue is related to [issue 92](https://github.com/yuanjian-org/app/issues/92)

Currently, if we are disconnected from database and the user tries to log in, the toast is not displayed properly when the backend server returns error code 500.
current

Now, the toast error message is displayed as intended.
modify

### DIFF
--- a/src/pages/auth/login.tsx
+++ b/src/pages/auth/login.tsx
@@ -53,8 +53,9 @@ export default function Login() {
         router.push(`/auth/verify`);
       }
     } catch (err) {
-      console.error(`Unkonwn error on sign-in: ${err}`);
-      toast.error(`糟糕，系统错误，请联系管理员：${err}`);
+      const msg = `糟糕，系统错误，请联系管理员：${err}`;
+      console.error(msg);
+      toast.error(msg);
     } finally {
       setIsLoading(false);
     }
--- a/src/pages/auth/login.tsx
+++ b/src/pages/auth/login.tsx
@@ -52,6 +52,9 @@ export default function Login() {
         localStorage.setItem(localStorageKeyForLoginEmail, email);
         router.push(`/auth/verify`);
       }
+    } catch (err) {
+      console.error(`Unkonwn error on sign-in: ${err}`);
+      toast.error(`糟糕，系统错误，请联系管理员：${err}`);
     } finally {
       setIsLoading(false);
     }


weihanwang:
shouldn't the error message look like "500 error" as opposed to "syntax error"?

Adagggggg:
> shouldn't the error message look like "500 error" as opposed to "syntax error"?

The "syntax error" is thrown by [res](https://github.com/yuanjian-org/app/blob/main/src/pages/auth/login.tsx#L45)`const res = await signIn('sendgrid', { email, callbackUrl, redirect: false });`, where `signIn` is from `next-auth/react`, 
Here's the source code:
```

export async function signIn<
  P extends RedirectableProviderType | undefined = undefined
>(
  provider?: LiteralUnion<
    P extends RedirectableProviderType
      ? P | BuiltInProviderType
      : BuiltInProviderType
  >,
  options?: SignInOptions,
  authorizationParams?: SignInAuthorizationParams
): Promise<
  P extends RedirectableProviderType ? SignInResponse | undefined : undefined
> {
  const { callbackUrl = window.location.href, redirect = true } = options ?? {}

  const baseUrl = apiBaseUrl(__NEXTAUTH)
  const providers = await getProviders()

  if (!providers) {
    window.location.href = `${baseUrl}/error`
    return
  }

  if (!provider || !(provider in providers)) {
    window.location.href = `${baseUrl}/signin?${new URLSearchParams({
      callbackUrl,
    })}`
    return
  }

  const isCredentials = providers[provider].type === "credentials"
  const isEmail = providers[provider].type === "email"
  const isSupportingReturn = isCredentials || isEmail

  const signInUrl = `${baseUrl}/${
    isCredentials ? "callback" : "signin"
  }/${provider}`

  const _signInUrl = `${signInUrl}${authorizationParams ? `?${new URLSearchParams(authorizationParams)}` : ""}`

  const res = await fetch(_signInUrl, {
    method: "post",
    headers: {
      "Content-Type": "application/x-www-form-urlencoded",
    },
    // @ts-expect-error
    body: new URLSearchParams({
      ...options,
      csrfToken: await getCsrfToken(),
      callbackUrl,
      json: true,
    }),
  })

  const data = await res.json()

  // TODO: Do not redirect for Credentials and Email providers by default in next major
  if (redirect || !isSupportingReturn) {
    const url = data.url ?? callbackUrl
    window.location.href = url
    // If url contains a hash, the browser does not reload the page. We reload manually
    if (url.includes("#")) window.location.reload()
    return
  }

  const error = new URL(data.url).searchParams.get("error")

  if (res.ok) {
    await __NEXTAUTH._getSession({ event: "storage" })
  }

  return {
    error,
    status: res.status,
    ok: res.ok,
    url: error ? null : data.url,
  } as any
}
```
The "syntax error" is from  `const data = await res.json()` due to empty response body. It doesn't handle the scenario when the error code is 500.

src/pages/auth/login.tsx
```diff
@@ -52,6 +52,9 @@ export default function Login() {
         localStorage.setItem(localStorageKeyForLoginEmail, email);
         router.push(`/auth/verify`);
       }
+    } catch (err) {
+      console.error(`Unkonwn error on sign-in: ${err}`);
+      toast.error(`糟糕，系统错误，请联系管理员：${err}`);
```
weihanwang, line 57: Can change to:

```
const msg = `糟糕，系统错误，请联系管理员：${err}`;
console.error(msg);
toast.error(msg):
```

So the messages in the console and the toaster are consistent.

weihanwang:
> > shouldn't the error message look like "500 error" as opposed to "syntax error"?
> 
> The "syntax error" is thrown by [res](https://github.com/yuanjian-org/app/blob/main/src/pages/auth/login.tsx#L45)`const res = await signIn('sendgrid', { email, callbackUrl, redirect: false });`, where `signIn` is from `next-auth/react`, 

Interesting! If you are interested consider contributing an PR to next-auth to fix the issue.

PR #309, opened 2024-06-05
bug fix 291 - make resource full height

xuan7678:
make iframe full height
Screenshot 2024-06-04 at 11 03 52 PM

### DIFF
--- a/src/pages/resources.tsx
+++ b/src/pages/resources.tsx
@@ -1,5 +1,6 @@
 import {
   Text,
+  Box
 } from '@chakra-ui/react';
 import React from 'react';
 import { useUserContext } from "../UserContext";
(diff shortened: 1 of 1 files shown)


src/pages/resources.tsx
```diff
@@ -12,6 +13,10 @@ export default fullPage(() => {
     return <Text>无权查看本页。</Text>;
   }
 
-  return <iframe src="https://f179b1fd0cd3453e9f34d95e95dc5f.super.site"
-    width="100%" height="700px" />;
+  return (
+      <Box h="calc(100vh - 60px)">
```
weihanwang, line 17: Use {`calc(100vh - ${topbarHeight})`} instead to avoid hardcoded constants.

xuan7678: addressed

weihanwang (approved):

PR #325, opened 2024-08-14
Add map api route

UncleBu1728:
Added `map.ts` under `map/route` containing the skeleton of the routers. This initial implementation includes the router structure without any functions implemented.

### DIFF
--- a/src/api/map/routes/map.ts
+++ b/src/api/map/routes/map.ts
@@ -1,32 +1,28 @@
 import { procedure, router } from "../../trpc";
 import { authUser } from "../../auth";
-import db from "../../database/db";
-import { z } from "zod";
-import { noPermissionError, notFoundError } from "../../errors";
-import { isPermitted } from "shared/Role";
 
 const get = procedure
   .use(authUser("Mentee"||"Mentor"))
-  .query(async ({ input: type }) =>
+  .query(async () =>
 {
 });
 
 
 const list = procedure
   .use(authUser("Mentee"||"Mentor"))
-  .query(async ({ input: type }) =>
+  .query(async () =>
 {
 });
 
 const save = procedure
 .use(authUser("Mentee"||"Mentor"))
-.query(async ({ input: type }) =>
+.query(async () =>
 {
 });
 
 const update = procedure
 .use(authUser("Mentee"||"Mentor"))
-.query(async ({ input: type }) =>
+.query(async () =>
 {
 });
 
--- a/src/api/map/routes/map.ts
+++ b/src/api/map/routes/map.ts
@@ -7,9 +7,8 @@ const list = procedure
   .use(authUser())
   .input(zLatitude)
   .output(z.array(zLandmark))
-  // eslint-disable-next-line no-unused-vars
   // eslint-disable-next-line require-await
-  .query(async ({ input }) =>
+  .query(async () =>
 {
   return [];
 });
--- a/src/api/map/routes/map.ts
+++ b/src/api/map/routes/map.ts
@@ -3,27 +3,11 @@
 import { procedure, router } from "../../trpc";
 import { authUser } from "../../auth";
 import { z } from "zod";
-import { zRoles } from "shared/Role";
-
-const zLandmark = z.object({
-  定义: z.string(),
-  经度: z.string(),
-  纬度: z.string(),
-  层级: z.array(z.string()),
-  相关地标: z.array(z.string()),
-  工具箱: z.string(),
-});
-
-//type Landmark = z.TypeOf<typeof zLandmark>;
+import { zLatitude, zLandmark } from "shared/Map";
 
 const list = procedure
   .use(authUser())
-  .input(z.object({
-    name: z.string(),
-    email: z.string(),
-    id: z.string(),
-    roles: zRoles,
-  }))
+  .input(zLatitude)
   .output(z.array(zLandmark))
   .query(async ({ input }) =>
 {
--- a/src/api/map/routes/map.ts
+++ b/src/api/map/routes/map.ts
@@ -1,14 +1,14 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
-/* eslint-disable require-await */
 import { procedure, router } from "../../trpc";
 import { authUser } from "../../auth";
 import { z } from "zod";
-import { zLatitudes, zLandmark } from "shared/Map";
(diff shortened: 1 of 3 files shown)


src/api/map/routes/map.ts
```diff
@@ -0,0 +1,34 @@
+import { procedure, router } from "../../trpc";
+import { authUser } from "../../auth";
+
+const get = procedure
+  .use(authUser("Mentee"||"Mentor"))
```
weihanwang, line 5: no need for "Mentee"||"Mentor". every logged-in user should be able to use the map.

src/api/map/routes/map.ts
```diff
@@ -0,0 +1,34 @@
...
+  .query(async () =>
+{
+});
+
+
+const list = procedure
```
weihanwang, line 11: let's start with this single API function and remove all others. let's implement map reading end-to-end first before moving on to updating etc.

src/api/map/routes/map.ts
```diff
@@ -0,0 +1,34 @@
...
+  .use(authUser("Mentee"||"Mentor"))
+  .query(async () =>
```
weihanwang, line 13: can you write out input parameters and return type?

UncleBu1728:
Vercel is showing errors due to the Eslint rules. Do we need another PR to disable these rules for now? Or work on more API implementation?
image

src/api/map/routes/map.ts
```diff
@@ -0,0 +1,16 @@
+import { procedure, router } from "../../trpc";
+import { authUser } from "../../auth";
+import { z } from "zod";
+
+const list = procedure
+  .use(authUser())
+  .input(z.string())
```
weihanwang, line 7: an enum type should be explicitly defined. refer to the Role type.

src/api/map/routes/map.ts
```diff
@@ -0,0 +1,16 @@
...
+  .output(z.array(z.object({})))
```
weihanwang, line 8: a struct type should be explicitly defined, with fields like ”定义“、“进阶” etc.

weihanwang (changes requested):

src/api/map/routes/map.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable require-await */
+import { procedure, router } from "../../trpc";
+import { authUser } from "../../auth";
+import { z } from "zod";
+import { zRoles } from "shared/Role";
+
+const zLandmark = z.object({
```
weihanwang, line 8: this needs to be in a folder referencable by the front end.

src/api/map/routes/map.ts
```diff
@@ -0,0 +1,35 @@
...
+  层级: z.array(z.string()),
+  相关地标: z.array(z.string()),
+  工具箱: z.string(),
+});
+
+//type Landmark = z.TypeOf<typeof zLandmark>;
+
+const list = procedure
+  .use(authUser())
+  .input(z.object({
```
weihanwang, line 21: the input doesn't need any of these fields. it instead needs lattitude which should be an enum.

src/api/map/routes/map.ts
```diff
@@ -0,0 +1,35 @@
...
+/* eslint-disable require-await */
+import { procedure, router } from "../../trpc";
+import { authUser } from "../../auth";
+import { z } from "zod";
+import { zRoles } from "shared/Role";
+
+const zLandmark = z.object({
+  定义: z.string(),
+  经度: z.string(),
+  纬度: z.string(),
```
weihanwang, line 11: These should be enums

src/api/map/routes/map.ts
```diff
@@ -0,0 +1,35 @@
...
+  层级: z.array(z.string()),
+  相关地标: z.array(z.string()),
+  工具箱: z.string(),
```
weihanwang, line 14: this should be optional().

src/api/map/routes/map.ts
```diff
@@ -0,0 +1,35 @@
...
+import { authUser } from "../../auth";
+import { z } from "zod";
+import { zRoles } from "shared/Role";
+
+const zLandmark = z.object({
+  定义: z.string(),
+  经度: z.string(),
+  纬度: z.string(),
+  层级: z.array(z.string()),
+  相关地标: z.array(z.string()),
```
weihanwang, line 13: this should be optional

src/shared/Map.ts
```diff
@@ -0,0 +1,28 @@
+import { z } from "zod";
+
+export const Longtitude = [
```
weihanwang, line 3: `Longtitudes`

src/shared/Map.ts
```diff
@@ -0,0 +1,28 @@
...
+    "情感",
+    "智慧",
+    "实践"
+] as const;
+
+export const zLongtitude = z.enum(Longtitude);
+
+export const Latitude = [
```
weihanwang, line 11: ditto

weihanwang (changes requested):

src/shared/Map.ts
```diff
@@ -1,25 +1,25 @@
...
 
-export const Longtitude = [
+export const Longtitudes = [
     "情感",
     "智慧",
     "实践"
 ] as const;
 
-export const zLongtitude = z.enum(Longtitude);
+export const zLongtitudes = z.enum(Longtitudes);
```
weihanwang, line 9: keep it as `zLongtitude`. singular.

src/shared/Map.ts
```diff
@@ -1,25 +1,25 @@
...
 
-export const Latitude = [
+export const Latitudes = [
     "个人成长",
     "事业发展",
     "社会责任"
 ] as const;
 
-export const zLatitude = z.enum(Latitude);
+export const zLatitudes = z.enum(Latitudes);
```
weihanwang, line 17: ditto.

src/api/map/routes/map.ts
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable require-await */
```
weihanwang, line 2: Use per-line statement and not per-file. Move them to immediately above the lines where lint rules are violated.

weihanwang (approved):

PR #343, opened 2024-09-05
Include file name in return array

UncleBu1728:
Include the landmark file names so the cards can be displayed correctly at the front end.


src/api/map/routes/map.ts
```diff
@@ -8,7 +8,7 @@ import * as path from 'path';
 const list = procedure
   .use(authUser())
   .input(zLatitude)
-  .output(z.array(zLandmark))
+  .output(z.array(zLandmark.extend({ fileName: z.string() })))
```
weihanwang, line 11: Don't extend the type here. Add a new field "名称“ to zLandmark. The code below should insert the value of this new field into the json object read from the file.

weihanwang (approved):

PR #352, opened 2024-09-08
Insert snippet at cursor position

zxcv5678:
Insert the snippet after clicking any position, and then the cursor will be set to the end of the current markdown.
Screenshot 2024-09-08 at 15 27 29

### DIFF
--- a/src/components/ChatRoom.tsx
+++ b/src/components/ChatRoom.tsx
@@ -10,8 +10,10 @@ import {
   VStack,
   Select,
   Link,
+  useBreakpointValue,
+  Flex,
 } from '@chakra-ui/react';
-import React, { useState } from 'react';
+import React, { useState, useRef } from 'react';
 import { ChatMessage } from 'shared/ChatMessage';
 import { componentSpacing, paragraphSpacing } from 'theme/metrics';
 import trpc, { trpcNext } from 'trpc';
@@ -103,9 +105,33 @@ function Editor({ roomId, message, onClose, ...rest }: {
     message ? message.markdown : "");
   const [saving, setSaving] = useState<boolean>(false);
   const utils = trpcNext.useContext();
-  
-  const insertSnippet = (snippet: string) => {
-    setMarkdown(prev => prev + snippet);
+  const direction: "row" | "column" | undefined = useBreakpointValue({
+    base: 'column' as const,
+    md: 'row' as const,
+  });
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
+  const insertSnippet = (markdown: string, cursorPos: number, snippet: string) => 
+  {
+    const before = markdown.substring(0, cursorPos);
+    const after = markdown.substring(cursorPos);
+    return before + snippet + after;
+  };
+
+  const handleSelectChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    const textarea = textareaRef.current;
+    if (textarea) {
+      const cursorPos = textarea.selectionStart; 
+      const snippet = snippets.find(snippet => 
+        snippet.text === e.target.value)?.text ?? '';
+      const updatedMarkdown = insertSnippet(markdown, cursorPos, snippet);
+      setMarkdown(updatedMarkdown); 
+
+      setTimeout(() => {
+        const newCursorPos = updatedMarkdown.length;
+        textarea.selectionStart = textarea.selectionEnd = newCursorPos;
+        textarea.focus(); 
+      }, 0);
+    }
   };
 
   const save = async () => {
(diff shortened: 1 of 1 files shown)


src/components/ChatRoom.tsx
```diff
@@ -126,39 +152,38 @@ function Editor({ roomId, message, onClose, ...rest }: {
...
     />
 
-    <HStack width="100%" spacing={componentSpacing}>
-      <Button onClick={save} isLoading={saving} isDisabled={!markdown}
-        variant="brand" leftIcon={<Icon as={MdSend} />}
-      >
-        确认
-      </Button>
-      <Button onClick={() => onClose()} variant="ghost" color="grey">取消</Button>
+    <Flex direction={direction} width="100%" gap={2}>
```
weihanwang, line 160: Avoid magic numbers

src/components/ChatRoom.tsx
```diff
@@ -103,9 +105,33 @@ function Editor({ roomId, message, onClose, ...rest }: {
     message ? message.markdown : "");
   const [saving, setSaving] = useState<boolean>(false);
   const utils = trpcNext.useContext();
-  
-  const insertSnippet = (snippet: string) => {
-    setMarkdown(prev => prev + snippet);
+  const direction: "row" | "column" | undefined = useBreakpointValue({
```
weihanwang, line 108: Why not at line 160, simply do `direction={{ base: "column", [sidebarBreakpoint]: "row" }}`?

src/components/ChatRoom.tsx
```diff
@@ -103,9 +105,33 @@ function Editor({ roomId, message, onClose, ...rest }: {
...
+  const handleSelectChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    const textarea = textareaRef.current;
+    if (textarea) {
+      const cursorPos = textarea.selectionStart; 
+      const snippet = snippets.find(snippet => 
+        snippet.text === e.target.value)?.text ?? '';
+      const updatedMarkdown = insertSnippet(markdown, cursorPos, snippet);
+      setMarkdown(updatedMarkdown); 
+
+      setTimeout(() => {
```
weihanwang, line 129: Add a comment on why using timeout

src/components/ChatRoom.tsx
```diff
@@ -103,9 +105,33 @@ function Editor({ roomId, message, onClose, ...rest }: {
...
+  const insertSnippet = (markdown: string, cursorPos: number, snippet: string) => 
+  {
+    const before = markdown.substring(0, cursorPos);
+    const after = markdown.substring(cursorPos);
+    return before + snippet + after;
+  };
+
+  const handleSelectChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    const textarea = textareaRef.current;
+    if (textarea) {
```
weihanwang, line 122: `if (!textarea) return` instead

src/components/ChatRoom.tsx
```diff
@@ -103,9 +105,33 @@ function Editor({ roomId, message, onClose, ...rest }: {
...
+      const cursorPos = textarea.selectionStart; 
+      const snippet = snippets.find(snippet => 
+        snippet.text === e.target.value)?.text ?? '';
+      const updatedMarkdown = insertSnippet(markdown, cursorPos, snippet);
+      setMarkdown(updatedMarkdown); 
+
+      setTimeout(() => {
+        const newCursorPos = updatedMarkdown.length;
```
weihanwang, line 130: Why do you need to update cursor position? Why can't leave its position as is?

src/components/ChatRoom.tsx
```diff
@@ -103,9 +105,33 @@ function Editor({ roomId, message, onClose, ...rest }: {
...
   const [saving, setSaving] = useState<boolean>(false);
   const utils = trpcNext.useContext();
-  
-  const insertSnippet = (snippet: string) => {
-    setMarkdown(prev => prev + snippet);
+  const direction: "row" | "column" | undefined = useBreakpointValue({
+    base: 'column' as const,
+    md: 'row' as const,
+  });
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
```
weihanwang, line 112: add an empty line below this line

weihanwang (changes requested):

src/components/ChatRoom.tsx
zxcv5678: The cursor position will be updated to where the user wants to insert the snippet, and then it will move to the end of the updated textarea so the user can continue typing.

weihanwang: If the user wants to insert the snippet at a certain position, it's more natural for the cursor stays at the end of the snippet than the end of the whole text, because it's more consistent with the cursor's expected behavior (right after the last inserted character).

src/components/ChatRoom.tsx
```diff
@@ -126,39 +153,38 @@ function Editor({ roomId, message, onClose, ...rest }: {
...
     />
 
-    <HStack width="100%" spacing={componentSpacing}>
-      <Button onClick={save} isLoading={saving} isDisabled={!markdown}
-        variant="brand" leftIcon={<Icon as={MdSend} />}
-      >
-        确认
-      </Button>
-      <Button onClick={() => onClose()} variant="ghost" color="grey">取消</Button>
+    <Flex direction={direction} width="100%" gap={mobileGap}>
```
weihanwang, line 161: What error would you see if you use `<Flex direction={{ base: "column", [sidebarBreakpoint]: "row" }} ...>` here?

zxcv5678: Screenshot 2024-09-08 at 17 12 54
Screenshot 2024-09-08 at 17 13 02

PR #355, opened 2024-09-11
Add landmark table

UncleBu1728:
Added `Landmark.ts` model under `api/database/models/map/`


src/api/database/models/map/Landmark.ts
```diff
@@ -0,0 +1,37 @@
...
+  Unique,
+  BelongsTo,
+} from "sequelize-typescript";
+import { UUID, STRING, JSONB, DATE } from "sequelize";
+import User from "../User";
+import ZodColumn from "../../modelHelpers/ZodColumn";
+import z from "zod";
+
+@Table
+class Landmark extends Model {
```
weihanwang, line 16: `LandmarkScore` instead.

src/api/database/models/map/Landmark.ts
```diff
@@ -0,0 +1,37 @@
...
+  @PrimaryKey
+  @Column(STRING)
+  landmark: string;
+
+  @Unique
+  @PrimaryKey
+  @ForeignKey(() => User)
+  @Column(UUID)
```
weihanwang, line 24: The primary key should be (userId, landmark, createdAt), in this order. createdAt is a built-in column in `Model`, so I believe you need to use class-level primary key definition instead of `@PrimaryKey` here.

src/api/database/models/map/Landmark.ts
```diff
@@ -0,0 +1,37 @@
...
+  userId: string;
+
+  @ZodColumn(JSONB, z.record(z.string(), z.any()).nullable())
+  evalutaion: Record<string, any> | null;
```
weihanwang, line 28: split into two fields:

`score: number` with Zod validation of range [1, MAX_LANDMARK_SCORE]. Define MAX_LANDMARK_SCORE in shared/Map.ts and replace the `4` in that file with this constant.

`markdown: string`. This is a note field where the user can jog down notes when scoring. The mock UI design doesn't have this field but we will need to add a text input box in the real UI.

src/api/database/models/map/Landmark.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  @Column(DATE)
+  history: string;
```
weihanwang, line 31: remove this.

src/api/database/models/map/Landmark.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  @BelongsTo(() => User)
+  user: User;
```
weihanwang, line 34: remove this unless you actually use it.

weihanwang (changes requested):

src/api/database/models/map/LandmarkScore.ts
```diff
@@ -0,0 +1,40 @@
...
+  @PrimaryKey
+  @ForeignKey(() => User)
+  @Column(UUID)
+  userId: string;
+
+  @PrimaryKey
+  @Column(STRING)
+  landmark: string;
+
+  @ZodColumn(INTEGER, z.number().int().min(1).max(MAX_LANDMARK_SCORE))
```
weihanwang, line 33: In shared/Map.ts, define a zod type `zLandmarkScore`. Use the type here.

src/api/database/models/map/LandmarkScore.ts
```diff
@@ -0,0 +1,40 @@
...
+  score: number;
```
weihanwang, line 34: use type `LandmarkScore` to be defined in `shared/Map.ts`

weihanwang: This comment is not addressed. It points to line 34 and not line 33.

src/api/database/models/map/LandmarkScore.ts
```diff
@@ -0,0 +1,39 @@
...
+@Table({
+  indexes: [
+    {
+      unique: true,
+      // Composite key on userId, landmark, createdAt
+      fields: ['userId', 'landmark', 'createdAt'],
+    },
+  ],
+})
+class LandmarkScore extends Model {
```
weihanwang, line 22: To avoid conflict with the type, rename to `LandmarkScoreLog`

src/api/database/models/map/LandmarkScore.ts
```diff
@@ -0,0 +1,39 @@
...
+  @ForeignKey(() => User)
+  @Column(UUID)
+  userId: string;
+
+  @PrimaryKey
+  @Column(STRING)
+  landmark: string;
+
+  @ZodColumn(INTEGER, zLandmarkScore)
+  score: Score;
```
weihanwang, line 33: Remove type aliasing

weihanwang (approved):

PR #370, opened 2024-09-23
try open in new window otherwise open in current window

### LINKED ISSUE #369: Auto detect pop-up blocker for Join Meeting button in GroupBar.tsx
Instead of using page sizes (breaking points) which is unreliable, always try to pop up the Tencent meeting page and fall back to opening in the same window if failed. See https://stackoverflow.com/a/2917

ada-only-one-ada:
fixed #369 
When the button is clicked, it should always attempt to open in a new window. If the pop-up window is blocked, then it should open in the current window.

### DIFF
--- a/src/components/GroupBar.tsx
+++ b/src/components/GroupBar.tsx
@@ -63,7 +63,7 @@ export default function GroupBar({
         const newWindow = window.open(link, '_blank');
         if (!newWindow || newWindow.closed || typeof newWindow.closed === 'undefined') {
           // popup blocked
-          void launchMeetingInSameWindow(groupId);
+          await launchMeetingInSameWindow(groupId);
         } else {
           setJoining(false);
         }
--- a/src/components/GroupBar.tsx
+++ b/src/components/GroupBar.tsx
@@ -61,9 +61,13 @@ export default function GroupBar({
         setShowMeetingQuotaWarning(true);
         setJoining(false);
       } else {
-        const newWindow = window.open(link, '_blank');
+        // Attempts to open the link in a new browser tab.
+        // If blocked or unsuccessful, it opens the link in the current tab.
+        // Ref: https://stackoverflow.com/a/2917 
+        const newWindow = window.open(link, '_blank');  
         if (!newWindow || newWindow.closed || typeof newWindow.closed === 'undefined') {
           window.location.href = link;
+          // Time is needed for the meeting page to load.
           setTimeout(() => setJoining(false), 5000);
         } else {
           setJoining(false);
--- a/src/components/GroupBar.tsx
+++ b/src/components/GroupBar.tsx
@@ -53,28 +53,7 @@ export default function GroupBar({
   const [isJoiningMeeting, setJoining] = useState(false);
   const [showMeetingQuotaWarning, setShowMeetingQuotaWarning] = useState(false);
 
-  const launchMeetingInNewWindow = async (groupId: string) => {
-    setJoining(true);
-    try {
-      const link = await trpc.meetings.join.mutate({ groupId: groupId });
-      if (!link) {
-        setShowMeetingQuotaWarning(true);
-      } else {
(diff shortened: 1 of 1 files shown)


src/components/GroupBar.tsx
```diff
@@ -60,10 +60,17 @@ export default function GroupBar({
       if (!link) {
         setShowMeetingQuotaWarning(true);
       } else {
-        window.open(link, '_blank');
+        const newWindow = window.open(link, '_blank');
+        if (!newWindow || newWindow.closed || typeof newWindow.closed === 'undefined') {
+          // popup blocked
+          await launchMeetingInSameWindow(groupId);
```
weihanwang, line 66: We shouldn't simply call launchMeetingInSameWindow() because calling the same RPC twice is wasteful and overall logic may not be correct. Merge these two functions instead. Note that their logic is similar but they have nuanced differences. Be very careful when change the code here. Make sure you fully understand the logic before making the change.

weihanwang (changes requested):

src/components/GroupBar.tsx
```diff
@@ -53,31 +53,21 @@ export default function GroupBar({
...
       const link = await trpc.meetings.join.mutate({ groupId: groupId });
       if (!link) {
         setShowMeetingQuotaWarning(true);
         setJoining(false);
       } else {
-        window.location.href = link;
-        // Time is needed for the meeting page to load.
-        setTimeout(() => setJoining(false), 5000);
+        const newWindow = window.open(link, '_blank');
+        if (!newWindow || newWindow.closed || typeof newWindow.closed === 'undefined') {
```
weihanwang, line 65: link to the reference page where you get this trick. Also add a comment about the intent of this block

src/components/GroupBar.tsx
```diff
@@ -53,31 +53,21 @@ export default function GroupBar({
...
+          window.location.href = link;
+          setTimeout(() => setJoining(false), 5000);
```
weihanwang, line 67: do not remove the original comment above this line.

src/components/GroupBar.tsx
```diff
@@ -53,31 +53,21 @@ export default function GroupBar({
...
-      } else {
-        window.open(link, '_blank');
-      }
-    } finally {
-      setJoining(false);
-    }
-  };
-
-  const launchMeetingInSameWindow = async (groupId: string) => {
+  const launchMeeting = async (groupId: string) => {
```
weihanwang, line 56: Test on all the following devices using the preview link of this PR to make sure it works as expected:

1. Desktop browsers: Chrome, Edge, Safari, Firefox
2. an iPhone
3. an Android phone

If you don't have all the devices, ask for volunteers who have to help you test.

ada-only-one-ada: ### iPhone Safari 

- **用户设置阻止弹窗时**
  - 自动在当前标签页打开腾讯会议。**Status: Pass**

- **用户设置不阻止弹窗时**
  - 弹出一个 Pop-up 窗口，提示："网站正在打开弹出式窗口"。
    - **用户选择允许：**
      - 在新标签页打开腾讯会议。**Status: Pass**
    - **用户选择阻止：**
      - 底部显示：```糟糕！Load failed```错误。见以下视频。
      - 在当前标签页打开腾讯会议后，底部错误消失。**Status: Unsure**

https://github.com/user-attachments/assets/8380ed85-f7e1-4a57-9bd5-0d94a73c49b2

### iPhone Chrome & iPhone Firefox

- **用户设置阻止弹窗时**
  - 自动在当前标签页打开腾讯会议。**Status: Pass**

- **用户设置不阻止弹窗时**
  - 自动在新标签页打开腾讯会议。**Status: Pass**

### iPhone Edge
- 无论用户是否设置阻止弹窗，都自动在新标签页打开腾讯会议。**Status: Pass**

weihanwang (approved):

PR #371, opened 2024-09-23
Add list landmarkAssessment api

UncleBu1728:
This PR adds ~~`listLandmarkScoreLog`~~ `listLandmarkAssessment` api to return ~~`LandmarkScoreLog`~~ `LandmarkAssessment` to the front end


src/shared/Map.ts
```diff
@@ -35,3 +35,20 @@ export type Landmark = z.TypeOf<typeof zLandmark>;
...
+    userId: z.string(),
+    landmark: z.string(),
+    createdAt: z.date(),
+    score: zLandmarkScore,
+    markdown: z.string().nullable(),
+});
+
+export type LandmarkScoreLog = z.TypeOf<typeof zLandmarkScoreLog>;
+
+export const landmarkScoreLogAttributes = [
```
weihanwang, line 49: Move it to attributesAndIncludes.ts. This is not supposed to be shared with the frontend.

src/shared/Map.ts
```diff
@@ -35,3 +35,20 @@ export type Landmark = z.TypeOf<typeof zLandmark>;
 export const zLandmarkScore = z.number().int().min(1).max(MAX_LANDMARK_SCORE);
 
 export type LandmarkScore = z.TypeOf<typeof zLandmarkScore>;
+
+export const zLandmarkScoreLog = z.object({
```
weihanwang, line 39: `zLandmarkScore`. "log" means an array of scores.

src/shared/Map.ts
```diff
@@ -35,3 +35,20 @@ export type Landmark = z.TypeOf<typeof zLandmark>;
...
+    userId: z.string(),
+    landmark: z.string(),
```
weihanwang, line 41: you don't need these two fields. They are input keys already.

src/shared/Map.ts
```diff
@@ -35,3 +35,20 @@ export type Landmark = z.TypeOf<typeof zLandmark>;
...
+    createdAt: z.date(),
+    score: zLandmarkScore,
+    markdown: z.string().nullable(),
+});
+
+export type LandmarkScoreLog = z.TypeOf<typeof zLandmarkScoreLog>;
+
+export const landmarkScoreLogAttributes = [
+    "userId",
+    "landmark",
```
weihanwang, line 51: remove these

src/api/map/routes/map.ts
```diff
@@ -28,6 +36,19 @@ const list = procedure
     }));
 });
 
+const listLandmarkScoreLog = procedure
+  .use(authUser())
+  .input(z.object({ userId: z.string(), landmark: z.string() }))
```
weihanwang, line 41: use a separate line for each field

weihanwang (approved):

weihanwang (changes requested):
need changes

src/shared/Map.ts
UncleBu1728: it would conflict with the `const zLandmarkScore = z.number().int().min(1).max(MAX_LANDMARK_SCORE)` above, should it be renamed to `zScore` and `zLandmarkScore`

weihanwang: I see. How about renaming this data type and also the table to LandmarkAssessment?

Also in a separate PR, add an assessorId field to the model (and this type), so that we will support assessors who are not the current user.

UncleBu1728:
TODO:  add accessorID field to allow permitted users to access.

src/api/database/models/attributesAndIncludes.ts
```diff
@@ -137,3 +137,13 @@ export const chatRoomInclude = [{
...
 }];
+
+/**
+ * Map
+ */
+
+export const landmarkAssessmentAttributes = [
+    "score", 
+    "markdown", 
+    "createdAt"
```
weihanwang, line 148: split into a new file: map/attributesAndIncludes.ts

src/api/map/routes/map.ts
```diff
@@ -28,6 +38,22 @@ const list = procedure
...
 });
 
+const listLandmarkAssessment = procedure
+  .use(authUser())
+  .input(z.object({ 
+    userId: z.string(), 
+    landmark: z.string() 
+  }))
+  .output(z.array(zLandmarkAssessment))
+  .query(async ({ input }) =>
```
weihanwang, line 48: `input` => `input: { userId, landmark }`

src/api/map/routes/map.ts
```diff
@@ -28,6 +38,22 @@ const list = procedure
...
+{
+  return await db.LandmarkAssessment.findAll({
+    where: { userId: input.userId, landmark: input.landmark },
+    attributes: landmarkAssessmentAttributes,
+  }) as LandmarkAssessment[];
```
weihanwang, line 53: you shouldn't need this `as` type casting.

UncleBu1728: if missing the `as` type casting, it would show an error as `Property 'createdAt' is optional in type 'LandmarkAssessment' but required in type '{ createdAt: Date; score: number; markdown: string | null; }'.`

weihanwang: add this comment to the code.

src/api/database/models/map/attributesAndIncludes.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Map
+ */
```
weihanwang, line 3: remove this comment block

weihanwang (dismissed):

weihanwang (approved):

PR #387, opened 2024-09-30
Add AssessorId to LandmarkAssessment Table

UncleBu1728:
Add `assessorId` in `LandmarkAssessment`to allow permitted users to access specific Landmark assessment histories.

### DIFF
--- a/src/api/database/models/User.ts
+++ b/src/api/database/models/User.ts
@@ -122,9 +122,6 @@ class User extends Model {
   @BelongsTo(() => User, { foreignKey: 'pointOfContactId' })
   pointOfContact: User | null;
 
-  @HasMany(() => LandmarkAssessment)
-  landmarkAssessments: LandmarkAssessment[];
-
   @BeforeDestroy
   static async cascadeDelete(user: User, options: any) {
 
--- a/src/api/database/models/User.ts
+++ b/src/api/database/models/User.ts
@@ -26,7 +26,6 @@ import Interview from "./Interview";
 import GroupUser from "./GroupUser";
 import Mentorship from "./Mentorship";
 import { MenteeStatus, zMenteeStatus } from "../../../shared/MenteeStatus";
-import LandmarkAssessment from "./map/LandmarkAssessment";
 
 
 @Table({ paranoid: true, tableName: "users", modelName: "user" })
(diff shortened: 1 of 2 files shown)


weihanwang (approved):

weihanwang (changes requested):
assessorId can't be null

UncleBu1728:
> assessorId can't be null

Does it mean we must assign a mentor to the record whenever we submit a landmark assessment?

weihanwang:
Yes. the assessor of self assessments is the current user.

weihanwang (dismissed):

UncleBu1728:
`@BelongsTo` is used instead of `@HasOne` in order to associate assessor with User model using assessorId. Otherwise, an error would be shown as below:

"throw new Error(`Foreign key for "${relatedClass.name}" is missing ` +
              ^Error: Foreign key for "LandmarkAssessment" is missing on "user"."

src/api/database/models/User.ts
```diff
@@ -121,6 +122,9 @@ class User extends Model {
   @BelongsTo(() => User, { foreignKey: 'pointOfContactId' })
   pointOfContact: User | null;
 
+  @HasMany(() => LandmarkAssessment)
+  landmarkAssessments: LandmarkAssessment[];
```
weihanwang, line 126: Don't add it for now. Let's add it only when we actual need it.

src/api/database/models/map/LandmarkAssessment.ts
```diff
@@ -23,6 +24,13 @@ class LandmarkAssessment extends Model {
   @Column(UUID)
   userId: string;
 
+  @ForeignKey(() => User)
+  @Column(UUID)
+  assessorId: string | null;
```
weihanwang, line 29: make a comment about what null means.

src/api/database/models/User.ts
```diff
@@ -26,6 +26,7 @@ import Interview from "./Interview";
 import GroupUser from "./GroupUser";
 import Mentorship from "./Mentorship";
 import { MenteeStatus, zMenteeStatus } from "../../../shared/MenteeStatus";
+import LandmarkAssessment from "./map/LandmarkAssessment";
```
weihanwang, line 29: remove this

src/api/database/models/map/LandmarkAssessment.ts
```diff
@@ -23,6 +24,15 @@ class LandmarkAssessment extends Model {
   @Column(UUID)
   userId: string;
 
+  // If assessorId (assessor) is null, the landmark assessment is self-evaluated by the user;
```
weihanwang, line 27: is this line > 80 chars long?

PR #395, opened 2024-10-09
Remove .env* files in docker image by adding it into dockerignore.

lijunray:


Dockerfile
```diff
@@ -67,6 +67,7 @@ USER nextjs
 EXPOSE 3000
 
 ENV PORT=3000
+ENV NEXT_PUBLIC_BRANDING=sizhu
```
weihanwang, line 70: Do not change content between "start of boilerplate" and "end of boilerplate" blocks

weihanwang (approved):

PR #399, opened 2024-10-11
make sidebar dragable to see user menu

### LINKED ISSUE #392: If the current window height is insufficient, unable to scroll down the sidebar.
Unable to scroll down the side bar:
Screenshot 2024-10-03 at 12 48 01

ada-only-one-ada:
Fixed #392.
If current window size is smaller, able to scroll down the side bar.

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -156,6 +156,7 @@ const Sidebar = ({ onClose, ...rest }: SidebarProps) => {
       <Flex
         direction="column"
         justifyContent="space-between"
+        overflowY="auto"
         h="full">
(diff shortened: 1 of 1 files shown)


src/components/Sidebar.tsx
```diff
@@ -156,6 +156,7 @@ const Sidebar = ({ onClose, ...rest }: SidebarProps) => {
       <Flex
         direction="column"
         justifyContent="space-between"
+        overflowY="auto"
```
weihanwang, line 159: Add a comment on why we need this

weihanwang (approved):

PR #417, opened 2024-10-17
change the column type for roles

### LINKED ISSUE #265: Use array instead of jsonb type for the user table’s roles column

xuan7678:
resolve #265

change the column type to array of varchar by following steps

1. create a new role with array type
2. copy all values from existing roles to the new role 
3. delete the old roles col
4. rename the new column to roles

### DIFF
--- a/src/api/database/models/User.ts
+++ b/src/api/database/models/User.ts
@@ -17,7 +17,7 @@ import {
   BelongsTo
 } from "sequelize-typescript";
 import Fix from "../modelHelpers/Fix";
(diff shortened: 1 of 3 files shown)


UncleBu1728:
Tested on both Windows and MacOS development environments, and the Role column type was successfully changed from JsonB to an Array without issues after running the migration script. The related functions and permissions seem to function as expected after modifying the role array. However, Ada encountered an issue during her development. Additional information is needed before proceeding with thorough testing.

PR #421, opened 2024-10-19
Ensure type checking 

ada-only-one-ada:


src/components/Sidebar.tsx
```diff
@@ -292,7 +292,9 @@ function DropdownMenu({ title, icon, menuItems, onClose } : {
             // Only sets the link it is a url 
             {...isUrl && { as: NextLink,  href: item.action } }
             onClick={() => {
-              (item as { action: Function }).action();
+              // (item as { action: Function }).action() 
+              // the above code will cause run time error if action is a URL.
+              if (typeof item.action === 'function') item.action();
```
weihanwang, line 297: The problem is not type conversion. The problem is the missing `if`. Change to:

```
if (!isUrl) (item.action as Function)();
```

src/components/Sidebar.tsx
```diff
@@ -292,7 +292,9 @@ function DropdownMenu({ title, icon, menuItems, onClose } : {
             // Only sets the link it is a url 
             {...isUrl && { as: NextLink,  href: item.action } }
             onClick={() => {
-              (item as { action: Function }).action();
+              // Ensure type checking before calling the function.
+              // Otherwise it will cause run time error if action is a URL.
```
weihanwang, line 296: With the new code the logic is obvious and self descriptive. the comment is no longer needed. it is outdated anyway.

weihanwang (approved):

PR #424, opened 2024-10-22
add getInterviewStats

### LINKED ISSUE #422: Create a route that lists interviewer stats
1. create a new route `getInterviewerStats` in file `src/api/routes/interviews.ts`
2. access to this route should be limited to MentorshipManagers only
3. It should return an array of users who has the role `Mentor`
4. The type of each array item should be like:
```
{
  user: zUser,
  interviews: z.number()
}
```
where `interviews` is the number of times the user is referred to in the InterviewFeedback.interviewerId field

ada-only-one-ada:
Fixed #422.

Screenshot 2024-10-22 at 11 41 15

### DIFF
--- a/src/api/routes/interviews.ts
+++ b/src/api/routes/interviews.ts
@@ -28,7 +28,6 @@ import { date2etag } from "./interviewFeedbacks";
 import { zFeedbackDeprecated } from "../../shared/InterviewFeedback";
 import { isPermittedForMentee } from "./users";
 import { zUser } from "../../shared/User";
-import { Op } from "sequelize";
 
 /**
  * Only MentorshipManager, interviewers of the interview, users allowed by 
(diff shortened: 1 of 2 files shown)


src/pages/interviews.tsx
weihanwang, line 1: Please undo changes in this file. Please create a separate page in the next PR, with its menu under "管理功能" as "面试统计“, accessible only to MentorshipManagers.

src/api/routes/interviews.ts
```diff
@@ -175,6 +182,34 @@ export async function createInterview(type: InterviewType, calibrationId: string
...
+.output(z.array(z.object({
+  user: zUser,
+  interviews: z.number(),
+})))
+.query(async () =>
+{
+  const mentors = await db.User.findAll({
+    attributes: userAttributes,
+    include: userInclude,
+    where: { roles: { [Op.contains]: ["Mentor"] }, }
```
weihanwang, line 196: Remove this line. We don't need to filter mentors.

src/api/routes/interviews.ts
```diff
@@ -175,6 +182,34 @@ export async function createInterview(type: InterviewType, calibrationId: string
...
+  });
+
+  const stats = await Promise.all(mentors.map(async mentor => {
```
weihanwang, line 199: It will be faster to read count all the interviewfeedback rows grouped by interviewids at once, and then compose the return result in memory.

weihanwang: the current code will issue many DB queries which can be very slow.

weihanwang (approved):

PR #434, opened 2024-10-25
show sex value in Chinese

ada-only-one-ada:
Currently display as 'female' or 'male', change it to empty/ 男 / 女


src/pages/interviewerStats.tsx
```diff
@@ -37,7 +37,7 @@ export default function Page() {
             <Tr key={user.id} _hover={{ bg: "white" }}> 
               <Td>{user.name}</Td>
               <Td>{interviewStat.interviews}</Td>
-              <Td>{user.sex}</Td>
+              <Td>{user.sex ? (user.sex === 'male' ? '男' : '女') : ''}</Td>
```
weihanwang, line 40: save "男“, "女" directly into the database rather than converting it in the frontend

Issue #441, opened 2024-10-26
Add partners section to /s/index.tsx


weihanwang:
Update <PartnersSection> in /s/index.tsx to add the following:

1. header: 参与机构
2. list all the logos and institution names that are listed at yuanjian.org/partners, excluding Tsinghua.
3. each GridItem should host one and only one institution. Create a reusable function to build these GridItems. Avoid repetitive code.

lijunray:
Should I download the images into public/img? Or should I just use the image link directly?

weihanwang:
> yuanjian.org/partners is blank for me, is it https://www.yjjxj.cn/partners?
> 
> Should I download the images into public/img? Or should I just use the image link directly?

download them into the repository.

PR #444, opened 2024-10-26
Add partners section to index page.

lijunray:
Screenshot 2024-10-25 at 10 12 14 PM

Styling could be refined later, but the basic data driven UI is built.

#441


src/pages/s/index.tsx
weihanwang, line 1: There is a build error. You can run `yarn build` locally to verify

src/pages/s/index.tsx
```diff
@@ -2,21 +2,40 @@ import { Grid, GridItem, GridProps, Heading, Text, VStack } from '@chakra-ui/rea
...
+}, {
+  name: "好奇学习社区", image: partner_5
+}];
 
 export default function Page() {
   return <VStack mt={50} spacing={50} align="start">
     <IntroSection />
     <ArticlesSection />
-    <PartnersSection />
+    <PartnersSection partners={partnersData} />
```
weihanwang, line 30: Don't change this line. move `partnersData` into PartnersSection function.

src/pages/s/index.tsx
```diff
@@ -70,6 +89,30 @@ function ArticlesSection() {
...
+
+function PartnersSection({ partners }: PartnersSectionProps) {
+  return <Flex gap={28} justify="flex-start" wrap="wrap">
+    {partners.map(partner => Partner(partner))}
+  </Flex>;
+}
+
+function Partner({ name, image }: Partner) {
+  return (
+    <Flex
```
weihanwang, line 109: Use GridItem instead, just like what other sections do. Use one GridItem per partner. This way the five partnership will be in a single row on desktop, and two partners per row on mobile, which is expected.

weihanwang: Use "参与机构" as the header parameter of the section

src/theme/metrics.ts
```diff
@@ -6,6 +6,8 @@ export const paragraphSpacing = 4;
 
 export const componentSpacing = 4;
 
+export const partnerSpacing = 10;
```
weihanwang, line 9: do not add it here. it's not going to be used by other components. inline this variable. just do `spacing={10}` in index.tsx:120

src/pages/s/index.tsx
```diff
@@ -1,8 +1,36 @@
...
 import wechatAppQrcode from '../../../public/img/wechat-app-qrcode.jpg';
+import partner_1 from '../../../public/img/partner_1.png';
+import partner_2 from '../../../public/img/partner_2.png';
+import partner_3 from '../../../public/img/partner_3.jpeg';
+import partner_4 from '../../../public/img/partner_4.png';
+import partner_5 from '../../../public/img/partner_5.png';
 import { sidebarBreakpoint } from 'components/Navbars';
+import { StaticImageData } from "next/image";
+
+const partnersData = [{
```
weihanwang, line 13: what I meant in this comment https://github.com/yuanjian-org/app/pull/444#discussion_r1817695712 is to move this data structure into the PartnersSection function. also change its name to simply `partners`.

src/pages/s/index.tsx
```diff
@@ -1,8 +1,36 @@
...
+  name: "中美爱心教育发展促进会",
+  image: partner_1,
+  link: "https://www.yjjxj.cn/blog/1d24bf123bb?categoryId=67479",
```
weihanwang, line 16: remove this links. we shouldn't link from this website to yuanjian's website.

src/pages/s/index.tsx
```diff
@@ -70,6 +98,33 @@ function ArticlesSection() {
   return <></>;
 }
 
+interface Partner {
```
weihanwang, line 101: no need for defining this type if `partnersData` is within the PartnersSection function.

weihanwang:
image the images should be centrally aligned vertically. the text should be also centrally aligned vertically.

lijunray:
> image the images should be centrally aligned vertically. the text should be also centrally aligned vertically.
> 
> text should be also centrally aligned horizontally. "中国科学技术大学" should be in a single line and not broken into two lines.

I don't know how to use existing `<Section>` which is based on `grid` to do this. I've tried a lot of different things but none worked. On the other hand, using `flex` will make this super easy. From my experience `grid` is a deprecated idea in frontend world after `flex` is out, especially for this case, the UI is simple enough that a flex can perfectly resolve all issues.

There is also not enough space to put all partners in the same line, unless we make the font very small, which is not scalable as well. Screenshot 2024-11-03 at 10 51 36 AM

weihanwang:
did you ask chatgpt? re: text on the same line, did you try to reduce spacing in between griditems?

lijunray:
> did you ask chatgpt? re: text on the same line, did you try to reduce spacing in between griditems?

I forced it to not wrap, but then the texts will exceed the border of each element.

PR #448, opened 2024-10-27
add role in interviewStats

### LINKED ISSUE #445: Add all Mentors, MentorCoaches & a new column to /interviewerStats table
Add all Mentors & MentorCoaches to the table. Each person should appear only once in the table.

The new column header is "角色“.

Column content of MentorCoaches should be like the following:
image

Column content of non-MentorCaoch Mentors should be like the following:
image

ada-only-one-ada:
Fixed #445 
Screenshot 2024-10-27 at 00 09 46

### DIFF
--- a/src/api/routes/interviews.ts
+++ b/src/api/routes/interviews.ts
@@ -28,7 +28,6 @@ import { date2etag } from "./interviewFeedbacks";
 import { zFeedbackDeprecated } from "../../shared/InterviewFeedback";
 import { isPermittedForMentee } from "./users";
 import { zUser } from "../../shared/User";
-import { Op } from "sequelize";
 
 /**
  * Only MentorshipManager, interviewers of the interview, users allowed by 
@@ -190,7 +189,7 @@ const getInterviewerStats = procedure
 })))
 .query(async () =>
 {
-  const interviewers = await db.User.findAll({
+  const users = await db.User.findAll({
     attributes: userAttributes,
     include: userInclude,
   });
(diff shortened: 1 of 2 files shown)


weihanwang:
place the role column right after name and before interview count columns.

src/api/routes/interviews.ts
```diff
@@ -189,7 +190,7 @@ const getInterviewerStats = procedure
 })))
 .query(async () =>
 {
-  const mentors = await db.User.findAll({
+  const interviewers = await db.User.findAll({
```
weihanwang, line 193: These are all the users. So should be named `users` instead

src/api/routes/interviews.ts
```diff
@@ -208,13 +209,29 @@ const getInterviewerStats = procedure
       return acc;
```
weihanwang, line 209: At line 206, comment on the structure of `interviewCountMap` (e.g. "a map of ....")

src/api/routes/interviews.ts
```diff
@@ -208,13 +209,29 @@ const getInterviewerStats = procedure
...
-    }))
-    .sort((i1, i2) => i1.interviews - i2.interviews);
+  const stats = interviewers
+    .filter(interviewer => interviewCountMap[interviewer.id])
+    .map(interviewer => ({
+      user: interviewer,
+      interviews: interviewCountMap[interviewer.id]
+    }));
+  
+  const mentors = await db.User.findAll({
```
weihanwang, line 219: Why do you need another query? You can already find all the mentors using the role field in the `users` list retrieved in 193. Simply modify the filter at line 213 to include mentors and mentorcoaches as well.

weihanwang (approved):

PR #450, opened 2024-10-28
set default wechat value if null

ada-only-one-ada:
After a user or a user manager update the wechat field value to empty, the invariant is checking not null and causing error in both 面试页 and申请页.

https://github.com/yuanjian-org/app/blob/52347819e494142fce24e165cc9f1aa810d4e195/src/components/Applicant.tsx#L109

### DIFF
--- a/src/api/routes/users.ts
+++ b/src/api/routes/users.ts
@@ -128,7 +128,7 @@ const update = procedure
   await user.update({
     name: input.name,
     city: input.city,
-    wechat: input.wechat || '（未提供微信）',
+    wechat: input.wechat,
     sex: input.sex,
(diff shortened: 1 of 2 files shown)


weihanwang:
Applicant.tsx:81 already ensures no empty value passed into <ContactFieldRow> so I don't know why this PR is necesary.

ada-only-one-ada:
> Applicant.tsx:81 already ensures no empty value passed into so I don't know why this PR is necesary.
Edited. 
 "" is a valid value, so the (??) won't replace it  because it only activates for null or undefined.

weihanwang (approved):

PR #471, opened 2024-11-08
351 better handling of tencentmeeting api errors

UncleBu1728:
Edit to show only "腾讯会议错误" instead of error code detail at the front end


src/api/TencentMeeting.ts
```diff
@@ -35,7 +35,8 @@ const requestWithBody = (body: string, options: {
     const req = (options.protocol === 'https:' ? https : http)
       .request(options, callback);
     req.on('error', (e: Error) => {
-      reject(e);
+      console.error('Request error:', e);
```
weihanwang, line 38: change to 'Tencent request error' to be more clear

src/api/TencentMeeting.ts
```diff
@@ -35,7 +35,8 @@ const requestWithBody = (body: string, options: {
...
+      reject(new Error('腾讯会议错误'));
```
weihanwang, line 39: I don't see how this change could result in better error message in the frontend. have you tested it?

I would change the following:

in errors.ts, add:

```
export const generalError = (message: string) =>
  new TRPCError({ code: 'INTERNAL_SERVER_ERROR', message });
```

then here:

```
throw generalError('腾讯会议后台错误，请重试');
```

please also make sure you test it out.

PR #528, opened 2024-11-27
correct readonly onClick

ada-only-one-ada:


weihanwang (approved):
fix spacing -- using Cursor

PR #561, opened 2024-12-02
use Flex to avoid overflow

### LINKED ISSUE #560: Fix text overflow issue on /volunteers cards
Example: image

ada-only-one-ada:
Fixed #560.
Screenshot 2024-12-02 at 13 04 00

### DIFF
--- a/src/components/UserCards.tsx
+++ b/src/components/UserCards.tsx
@@ -17,7 +17,7 @@ import {
 } from '@chakra-ui/react';
 import Loader from 'components/Loader';
 import { formatUserName, toPinyin } from 'shared/strings';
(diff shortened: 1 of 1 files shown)


src/components/UserCards.tsx
```diff
@@ -196,7 +197,7 @@ function UserCardForDesktop({ user, profile: p, type, openModal }: {
       </Heading>
     </CardHeader>
     <CardBody pt={1}>
-      <VStack align="start">
+      <Flex direction="column" gap="8px">
```
weihanwang, line 200: why not `gap={paragraphSpacing}`? avoid magic numbers as much as you can

weihanwang (approved):

PR #571, opened 2024-12-08
fix wechat login callback and add loading animation for QR

xingcan-hu:


src/components/WeChatQRLogin.tsx
```diff
@@ -11,29 +11,56 @@ declare global {
...
   const router = useRouter();
-  const callbackUrl = router.query.callbackUrl as string || '/';
+  const [isLoading, setIsLoading] = useState(true);
+
+  // 获取当前路径作为回调URL
+  const currentPath = typeof window !== 'undefined'
+    ? window.location.pathname + window.location.search
+    : '/';
+  const callbackUrl = router.query.callbackUrl as string ||
+    (currentPath !== '/auth/login' ? currentPath : '/');
```
weihanwang, line 21: Can you pass in the callbackUrl from the caller (ie. add a callbackUrl parameter to the WeChatQRLogin() function, instead of recomputing it here?

src/components/WeChatQRLogin.tsx
```diff
@@ -44,11 +71,47 @@ export default function WeChatQRLogin({ appid }: { appid: string }) {
...
+        top="0"
+        left="0"
+        width="300px"
+        height="170px"
+        display="flex"
+        flexDirection="row"
+        alignItems="center"
+        justifyContent="center"
+        bg="white"
+        zIndex="1"
```
weihanwang, line 86: do we need all these properties?

xingcan-hu: bg zIndex  removed.

PR #574, opened 2024-12-09
fix wechat QR login state CSRF issue

xingcan-hu:


src/pages/api/auth/[...nextauth].ts
```diff
@@ -39,90 +40,108 @@ export const adapter = SequelizeAdapter(sequelize, {
...
+  
+    pages: {
+      signIn: '/auth/login',
+      // The login page respects the `?error=` URL param.
+      error: '/auth/login',
+    },
+  
+    callbacks: {
+      signIn: ({ account }) => {
+        // https://github.com/nextauthjs/next-auth/discussions/469
```
Comment (xingcan-hu) [inline src/pages/api/auth/[...nextauth].ts:85]: add signIn

src/pages/api/auth/[...nextauth].ts
```diff
@@ -39,90 +40,108 @@ export const adapter = SequelizeAdapter(sequelize, {
...
+        id: 'sendgrid',
+        type: 'email',
+        maxAge: tokenMaxAgeInMins * 60, // For verification token expiry
+        sendVerificationRequest,
+        generateVerificationToken: generateShortLivedToken,
+      },
+      WeChatProvider({
+        id: "wechat-qr",
+        name: "微信扫码登陆",
+        checks: ["none"],
```
Comment (weihanwang) [inline src/pages/api/auth/[...nextauth].ts:63]: Do we still need this line here?

Comment (xingcan-hu) [reply src/pages/api/auth/[...nextauth].ts:65]: // The "checks: ['none']" is necessary because the current QR code login does not follow the full NA auth flow, 
        // and the backend does not generate a state. Removing this would cause the check to fail, preventing login.

weihanwang:
Merged as https://github.com/yuanjian-org/app/pull/583

